PR #8, opened 2022-05-18
fix no notification sound in Secondary user

empratyush:
closes [ #483](https://github.com/GrapheneOS/os-issue-tracker/issues/483)


thestinger:
Shouldn't this use `DEFAULT_LIGHTS | DEFAULT_SOUND`? It's disabling lights this way.

empratyush:
fixed now.

PR #16, opened 2023-02-14
[Bug fixes] add checks before casting

empratyush:
fixes [#1882](https://github.com/GrapheneOS/os-issue-tracker/issues/1882)


src/com/android/messaging/ui/conversationlist/ForwardMessageActivity.java
```diff
@@ -46,7 +46,9 @@ protected void onCreate(final Bundle savedInstanceState) {
 
     @Override
     public void onAttachFragment(final Fragment fragment) {
-        Assert.isTrue(fragment instanceof ConversationListFragment);
+        if (fragment == null || !(fragment instanceof ConversationListFragment)) {
```
muhomorr, line 49: `instanceof` check is always false when object is null.

PR #18, opened 2023-09-18
[bugfixes] handle null case

empratyush:
closes [#2424](https://github.com/GrapheneOS/os-issue-tracker/issues/2424)

### DIFF
--- a/src/com/android/messaging/datamodel/ParticipantRefresh.java
+++ b/src/com/android/messaging/datamodel/ParticipantRefresh.java
@@ -46,6 +46,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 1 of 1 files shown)


src/com/android/messaging/datamodel/ParticipantRefresh.java
```diff
@@ -390,8 +390,9 @@ private static void refreshSelfParticipantList() {
             // For subscriptions already in the database, refresh ParticipantColumns.SIM_SLOT_ID.
             for (final Integer subId : activeSubscriptionIdToRecordMap.keySet()) {
                 final SubscriptionInfo record = activeSubscriptionIdToRecordMap.get(subId);
-                final String displayName =
-                        DatabaseUtils.sqlEscapeString(record.getDisplayName().toString());
+                if (record == null) continue;
```
thestinger, line 393: This `record == null` check isn't needed since it's created above and the entries in the map can't be `null`.

src/com/android/messaging/datamodel/ParticipantRefresh.java
```diff
@@ -390,8 +390,8 @@ private static void refreshSelfParticipantList() {
             // For subscriptions already in the database, refresh ParticipantColumns.SIM_SLOT_ID.
             for (final Integer subId : activeSubscriptionIdToRecordMap.keySet()) {
                 final SubscriptionInfo record = activeSubscriptionIdToRecordMap.get(subId);
-                final String displayName =
-                        DatabaseUtils.sqlEscapeString(record.getDisplayName().toString());
```
thestinger, line 394: You can use `Objects.toString(record.getDisplayName(), "")` instead and just keep it on one line.

https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#toString-java.lang.Object-java.lang.String-

PR #26, opened 2024-12-23
Update to SDK 35

t895:
This has the migrations needed to bump `targetSdk` from 24 to 35


t895:
Updated this because I wasn't specifying the correct mutability on each PendingIntent. They should only be mutable when using bubbles or inline replies and this app uses neither of those.

muhomorr:
> Use READ_MEDIA permissions 
> Required for Android 14

Those permissions are needed for targetSdk 33 (Android 13).

muhomorr:
"Settings > Default SMS app" item doesn't work. Similar action works in Google Messages, it open an SMS role holder selection dialog.

t895:
Fixed both

PR #27, opened 2024-12-29
Replace FrameSequence with Glide

t895:


muhomorr:
Need to update CMakeLists.txt as well.

t895:
Removed

muhomorr:
CMakeLists.txt can't be removed yet, libgiftranscode is still needed.

t895:
Sorry about that, fixed

PR #30, opened 2025-01-04
fixup: PendingIntent Mutability and revert adding missing super calls in CameraMediaChooserView

t895:
- Fixes an issue where you can send MMS messages, but the app would say that it didn't send properly
- Revert "add missing super calls in CameraMediaChooserView."

These super calls seemingly were omitted intentionally since the framework throws an exception due to having multiple Views with the ID, NO_ID, if you invoke the attachments menu twice with the super calls included. I've just suppressed the warning as this doesn't cause any problems.


t895:
Removed the change to PendingIntent mutability since those were addressed in main

muhomorr:
I've intentionally didn't include the fix for `MmsSender#sendMms` issue (https://github.com/t895/platform_packages_apps_Messaging/commit/9374f5f01c4b4c9ffbd0f8a0ff6a31f84d9e2051) since it was already fixed in your PR.

I've submitted that commit in a separate PR: https://github.com/GrapheneOS/platform_packages_apps_Messaging/pull/32

t895:
Oh whoops
Thank you

PR #36, opened 2025-02-19
Prevent conversation deleted toast spam

t895:
Make the DeleteConversationAction capable of deleting multiple conversations at once and notify the user once it has deleted every conversation given to it

Previously it would make a toast message for every conversation it deleted. The toast message would then block a portion of the screen and interrupt the user for a significant period of time.

https://github.com/user-attachments/assets/a096dab8-819d-4a17-a54c-30f9b9a51a8f


mags0ft (approved):
Looks good, reasonable fix. There's no need to send that many toasts.

PR #76, opened 2025-08-07
Actually bump the minimum and target SDK to 36, since it seems to have been forgotten

andromxda:
Not incrementing the version number, since no changes have been made since the increment to 13, and version 13 hasn't been released yet


RankoR:
Will merge after completing transition to Compose. In the current state it’s not worth to fix edge-to-edge problems that it will cause in the legacy code.

RankoR:
Will bump to 37 when will complete the transition.

PR #78, opened 2025-08-17
Fix widget crashing 

Inhishonor:
Every time the user would add a widget, the app would crash, due to the receiver not being properly registered in onEnabled, causing a security exception.


src/com/android/messaging/widget/BaseWidgetProvider.java
```diff
@@ -81,7 +81,14 @@ public void onReceive(Context context, Intent intent) {
...
     public void onEnabled(Context context) {
         super.onEnabled(context);
-        context.getApplicationContext().registerReceiver(this, new IntentFilter(getAction()));
+        IntentFilter filter = new IntentFilter(getAction());
+        context.getApplicationContext().registerReceiver(
+                this,
+                filter,
+                null,
+                null,
+                Context.RECEIVER_EXPORTED
```
soupslurpr, line 90: Why use `Context.RECEIVER_EXPORTED` and not `Context.RECEIVER_NOT_EXPORTED`?

soupslurpr (changes requested):

src/com/android/messaging/widget/BaseWidgetProvider.java
Inhishonor: That was my first thought, but I wasn't sure. Do you want me to change it to `android:exported: false` etc.?

Inhishonor: Ok, that is done. I read through the code, and tested it pretty extensively, and everything seemed to work as before. Should be good to go.

PR #80, opened 2025-08-21
Add a warning about sms/mms insecurity

### LINKED ISSUE #47: Add a warning that sms is not safe
Add a warning that sms is not safe

Inhishonor:
Right now, shows a screen at app launch on each new version, with a welcome heading, and warning about SMS/MMS.
The only thing I am not sure about if showing it after every version would be to annoying for the user. I considered putting it as a dialog/snackbar/card, but couldn't figuring out a way that would be noticeable but not annoying. 
Closes #47

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -90,6 +90,11 @@
             android:screenOrientation="portrait"
             android:configChanges="orientation|screenSize|keyboardHidden" />
 
+        <activity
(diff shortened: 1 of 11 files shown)


RankoR:
Thank you for your contribution, but this PR will not be merged because we're currently migrating this app to Compose, and it's using legacy Views. We will consider adding this feature rewritten with Compose, or you can re-implement it with Compose if you wish (but a bit later, when we will finish the migration).

PR #85, opened 2025-10-19
Support dynamic colors

sizzlesloth:
Ideally I wanted to make the least changes necessary to support dynamic colors, but also keep what exists as a fallback.

Note:
I intended to change the most visible front-facing colors, keeping minor details (SMS error text, etc.) the defaults. I have no artistic talent whatsoever, so feel free to change the coloring of minor details if you wish.

### Changes
* Add `com.google.android.material:material:1.13.0` dep
* Add a `values-v31/colors.xml` file using token colors
* Renamed some icon size attributes to amend naming conflicts with new dep

### Out of scope
* Changing components from appcompat to material3
* Changing graphical asset colors (for example, the 'selection tick' icon is still green, when long-pressing on a conversation list item)


RankoR:
We're currently rewriting this app with Compose, so this PR isn't relevant anymore, but thanks for your contribution.

PR #101, opened 2026-04-03
Settings Screen: Java/XML to Kotlin/Compose migration + tests

m4pl:
Migrated the Settings screen from Java + XML to Kotlin + Compose. Added unit tests for Composable(s) and ViewModel. Additionally, wrote an E2E test covering the flow: user opens Conversations list screen and navigates to a conversation.

### Screenshots

| Before | After |
|--------|-------|
|Screenshot_20260403-202830|Screenshot_20260403-202659|
|Screenshot_20260403-202834|Screenshot_20260403-202702|
|Screenshot_20260403-202838|Screenshot_20260403-202706|

### Figma
https://figma.com/design/JtQuS3wQQs9YdwrtZtGw0y


src/com/android/messaging/di/settings/SettingsBindsModule.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.android.messaging.ui.appsettings.subscription.mapper.SubscriptionSettingsUiStateMapper
+import com.android.messaging.ui.appsettings.subscription.mapper.SubscriptionSettingsUiStateMapperImpl
+import dagger.Binds
+import dagger.Module
+import dagger.Reusable
+import dagger.hilt.InstallIn
+import dagger.hilt.components.SingletonComponent
+
+@Module
+@InstallIn(SingletonComponent::class)
```
RankoR, line 20: Should be `ViewModelComponent` probably

m4pl: Right, will fix

src/com/android/messaging/ui/appsettings/screen/SettingsScreen.kt
```diff
@@ -0,0 +1,178 @@
...
+                val sub = uiState.subscriptionSettings.find { it.subId == route.subId }
+                if (sub != null) {
+                    SubscriptionSettingsScreen(
+                        subscriptionSettings = sub,
+                        title = route.title,
+                        screenModel = screenModel,
+                        onNavigateBack = navigateUp,
+                    )
+                }
+            }
```
inthewaves, line 170: On a dual-SIM setup, if looking at settings for a SIM and that SIM is removed (eSIM is disabled or physical SIM is removed), sub will be null. Since nothing is used for `sub == null` case, the entire screen turns white until back gesture. 

We should show something (e.g. make `SubscriptionSettingsScreen` accept a nullable subscriptionSettings and handle it inside) or force the user to go back when SIM is removed (i.e., when `sub` is null) while SubscriptionSettingsScreen is currently being used (and then add a test for this case)

src/com/android/messaging/ui/appsettings/subscription/mapper/SubscriptionSettingsUiStateMapper.kt
```diff
@@ -0,0 +1,156 @@
...
+
+        val phoneNumberKey = context.getString(R.string.mms_phone_number_pref_key)
+        val savedPhoneNumber = subPrefs.getString(phoneNumberKey, "")
+        val defaultPhoneNumber = phoneUtils.getCanonicalForSelf(false)
+
+        val displayPhoneNumber = when {
+            !savedPhoneNumber.isNullOrEmpty() -> phoneUtils.formatForDisplay(savedPhoneNumber)
+            !defaultPhoneNumber.isNullOrEmpty() -> phoneUtils.formatForDisplay(defaultPhoneNumber)
+            else -> context.getString(R.string.unknown_phone_number_pref_display_value)
+        }
```
inthewaves: Current code only uses the equivalent of `PhoneUtils.get(mSubId).formatForDisplay(value)`

Previous code used `PhoneUtils.get(mSubId).formatForDisplay(value)` then `bidiFormatter.unicodeWrap(displayValue, TextDirectionHeuristicsCompat.LTR)`; for phone number preference summary.

https://github.com/GrapheneOS/Messaging/blob/f1e0efc2443d195380f5b2d656bb94d0151f93df/src/com/android/messaging/ui/appsettings/PhoneNumberPreference.java#L90-L105

It seems we should add this back. There's also the Compose-equivalent of this like `TextStyle(textDirection = TextDirection.ContentOrLtr)` for phone number `Text` or `OutlinedTextField`s. The BidiFormatter usage seems easier/more unconditional though + it would help if string is embedded somewhere else.

src/com/android/messaging/ui/appsettings/common/SettingsComponents.kt
```diff
@@ -0,0 +1,245 @@
...
+        summary = summary,
+        icon = icon,
+        enabled = enabled,
+        verticalPadding = 12.dp,
+        trailing = {
+            Spacer(modifier = Modifier.width(16.dp))
+            Switch(
+                checked = checked,
+                onCheckedChange = null,
+            )
```
inthewaves, line 116: the `enabled` flag should be passed to the switch. Currently the switch is not dimmed when the `SettingsSwitchItem`'s `enabled` is false

e.g., disable the Auto-retrieve preference and look at the switch for Roaming auto-retrieve, especially when Roaming auto-retrieve is on

src/com/android/messaging/ui/appsettings/screen/SettingsMainScreen.kt
```diff
@@ -0,0 +1,75 @@
...
+            contentPadding = contentPadding,
+        ) {
+            item {
+                SettingsClickableItem(
+                    title = stringResource(R.string.general_settings),
+                    onClick = onGeneralSettingsClick,
+                )
+            }
+
+            items(subscriptions) { subscription ->
```
inthewaves:
```suggestion
            items(subscriptions, key = { it.subId }) { subscription ->
```

app/src/androidTest/java/com/android/messaging/ui/conversation/ConversationUserFlowTest.kt
```diff
@@ -0,0 +1,76 @@
...
+        val instrumentation = InstrumentationRegistry.getInstrumentation()
+        val packageName = instrumentation.targetContext.packageName
+        val command = "cmd role add-role-holder android.app.role.SMS $packageName"
+        val parcelFileDescriptor = instrumentation.uiAutomation.executeShellCommand(command)
+
+        ParcelFileDescriptor.AutoCloseInputStream(parcelFileDescriptor).use { inputStream ->
+            val result = String(inputStream.readBytes())
+            println("Role assignment result: $result")
+        }
+    }
```
inthewaves: This seems like it could be in a `@BeforeClass` as well (if more tests are added, it'll run for every test)

inthewaves (changes requested):

AndroidManifest.xml
```diff
@@ -164,40 +164,17 @@
...
-            android:screenOrientation="user"
-            android:parentActivityName="com.android.messaging.ui.appsettings.SettingsActivity">
-            <meta-data
-                android:name="android.support.PARENT_ACTIVITY"
-                android:value="com.android.messaging.ui.appsettings.SettingsActivity" />
-        </activity>
-
-        <activity
-            android:name=".ui.appsettings.ApplicationSettingsActivity"
             android:exported="true"
```
inthewaves, line 169: Technically, this is removing the exported Activity `.ui.appsettings.ApplicationSettingsActivity` and just using `.ui.appsettings.SettingsActivity` now. 

Since `.ui.appsettings.ApplicationSettingsActivity` used to be exported, for completeness, maybe an `activity-alias` could be added here for `.ui.appsettings.ApplicationSettingsActivity` with `android:targetActivity=".ui.appsettings.SettingsActivity">`

src/com/android/messaging/ui/appsettings/screen/SettingsScreen.kt
m4pl: Good catch, this definitely needs to be handled

src/com/android/messaging/ui/appsettings/subscription/mapper/SubscriptionSettingsUiStateMapper.kt
m4pl: Thanks, I’ll take another look

src/com/android/messaging/ui/appsettings/common/SettingsComponents.kt
m4pl: Agreed, I’ll update it so it is less confusing for users

app/src/androidTest/java/com/android/messaging/ui/conversation/ConversationUserFlowTest.kt
m4pl: Yes, that is a good idea for future tests. I’ll update it

AndroidManifest.xml
m4pl: I will take a look, thanks

src/com/android/messaging/ui/appsettings/screen/SettingsScreen.kt
m4pl: I've updated the logic, so now you can disable all SIM cards or enable/disable a specific one, and the screen will automatically update via [OnSubscriptionsChangedListener](https://github.com/GrapheneOS/Messaging/pull/101/changes#diff-93d21617ededd675e92a750116236a4d95226f6f963f65941d78cf300fd64500R64).

src/com/android/messaging/ui/appsettings/ApplicationSettingsActivity.java
```diff
@@ -1,223 +0,0 @@
...
-                // Hide the Advanced settings screen if this is not top-level; these are shown at
-                // the parent SettingsActivity.
-                getPreferenceScreen().removePreference(advancedScreen);
-            }
-
-            final PreferenceScreen smsDisabled = (PreferenceScreen) findPreference(mSmsDisabledPrefKey);
-            smsDisabled.setOnPreferenceClickListener(preference -> {
-                RoleManager roleManager = getContext().getSystemService(RoleManager.class);
-                Intent intent = roleManager.createRequestRoleIntent(RoleManager.ROLE_SMS);
-                startActivityForResult(intent, 0);
```
inthewaves, line 144: Old role request code correctly uses `startActivityForResult`

src/com/android/messaging/ui/appsettings/screen/SettingsEffectHandler.kt
```diff
@@ -0,0 +1,57 @@
...
+                    putExtra(Settings.EXTRA_APP_PACKAGE, context.packageName)
+                }
+                context.startActivity(intent)
+            }
+
+            is SettingsScreenEffect.RequestDefaultSmsApp -> {
+                val roleManager = context.getSystemService(RoleManager::class.java)
+                val intent = roleManager.createRequestRoleIntent(RoleManager.ROLE_SMS)
+                context.startActivity(intent)
+            }
```
inthewaves: The documentation for `RoleManager#createRequestRoleIntent` says:

> Returns an Intent suitable for passing to `android.app.Activity.startActivityForResult(Intent, int)` which prompts the user to grant a role to this application.

`startActivityForResult` appears to be a more severe requirement that what's stated in that documentation, as using `startActivity` will result in the role request being ignored: https://cs.android.com/android/platform/superproject/+/android-latest-release:packages/modules/Permission/PermissionController/src/com/android/permissioncontroller/role/ui/RequestRoleActivity.java;l=94. Internally, using `startActivity` and not `startActivityForResult` appears to not set a `resultRecord`, so `getCallingPackage()` becomes null. 

logcat line `RequestRoleActivity     com.android.permissioncontroller     W  Package name cannot be null or empty: null` from PermissionController code can be seen when `startActivity` is used

```suggestion
            is SettingsScreenEffect.RequestDefaultSmsApp -> {
                val roleManager = context.getSystemService(RoleManager::class.java)
                val intent = roleManager.createRequestRoleIntent(RoleManager.ROLE_SMS)
                context.startActivityForResult(intent, 0)
            }
```

src/com/android/messaging/ui/appsettings/screen/SettingsScreen.kt
```diff
@@ -0,0 +1,215 @@
...
+    LaunchedEffect(screenModel, effectHandler) {
+        screenModel.effects.collect(effectHandler::handle)
+    }
+
+    // For single-SIM go directly to app settings
+    val effectiveRoute = if (uiState.isMultiSim == false && currentRoute is SettingsNavRoute.Main) {
+        SettingsNavRoute.AppSettings
+    } else {
+        currentRoute
+    }
```
inthewaves, line 72: Appears to regress single-SIM, I can see `SettingsNavRoute.Main` / `SettingsMainScreen` on single-SIM device before it quickly switches to `SettingsNavRoute.AppSettings` / `AppSettingsScreen`. Perhaps this needs to be computed synchronously in `SettingsActivity`

https://github.com/user-attachments/assets/9023ff5a-7542-4404-872e-ba06d5ad20fc

src/com/android/messaging/ui/appsettings/screen/SettingsScreen.kt
```diff
@@ -0,0 +1,215 @@
...
+    modifier: Modifier = Modifier,
+    initialRoute: SettingsNavRoute = SettingsNavRoute.Main,
+    screenModel: SettingsScreenModel = viewModel<SettingsViewModel>(),
+) {
+    val context = LocalContext.current
+    val uiState by screenModel.uiState.collectAsStateWithLifecycle()
+
+    var currentRoute by remember {
+        mutableStateOf(initialRoute)
+    }
```
inthewaves: Currently this doesn't survive system-initiated process death.

e.g., We can go to OS Settings > System > Developer options > Don't keep activities to simulate Android destroying the activity normally and needing to recreate it after process death. 

Then if we open a specific SIM setting or the General page for multi-SIM, pressing home button to exit the app and cause system to kill the app process. Then relaunching the app from recents will recreate the Activity. Since it's not saved, it kicks the user back to the SettingsMainScreen instead of keeping that subscription setting or general settings open. This didn't occur in the old Messages app settings code

We could implement `rememberSaveable` which would need a `Saver` for `SettingsNavRoute`. Note that this probably also needs to change the default value for multi SIM

src/com/android/messaging/ui/appsettings/subscription/delegate/SubscriptionSettingsDelegate.kt
```diff
@@ -0,0 +1,132 @@
...
+import kotlinx.coroutines.channels.awaitClose
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.callbackFlow
+import kotlinx.coroutines.flow.drop
+import kotlinx.coroutines.launch
+
+internal data class SubscriptionSettingsState(
+    val isMultiSim: Boolean = false,
```
inthewaves: I think it's better to make this val type nullable and make the default null, especially if we do `rememberSaveable` for the route in `SettingsScreen`. We shouldn't assume not multi SIM for initial states

inthewaves (changes requested):

src/com/android/messaging/ui/appsettings/screen/model/SettingsUiState.kt
```diff
@@ -0,0 +1,12 @@
...
+import androidx.compose.runtime.Immutable
+import com.android.messaging.ui.appsettings.general.model.AppSettingsUiState
+import com.android.messaging.ui.appsettings.subscription.model.SubscriptionSettingsUiState
+
+@Immutable
+internal data class SettingsUiState(
+    val isMultiSim: Boolean? = null,
+    val subscriptionSettings: List<SubscriptionSettingsUiState> = emptyList(),
+    val appSettings: AppSettingsUiState = AppSettingsUiState(),
+)
```
inthewaves, line 15: Just as a comment: Kotlin `List` isn't technically immutable; it just doesn't have mutating methods in the interface. The backing list could be `ArrayList`. In practice, I think the `List`s passed to the constructor of `SettingsUiState` are never mutated and is always kept to the Kotlin `List` interface (via `listOf` or `buildList`), so it's never exposed as an `ArrayList` or something with mutable methods

In the future there's https://github.com/Kotlin/kotlinx.collections.immutable if we want guarantees but not worth the effort right now I think

src/com/android/messaging/ui/appsettings/screen/SettingsScreen.kt
```diff
@@ -0,0 +1,215 @@
...
+private fun rememberDisplayedSubscription(
+    route: SettingsNavRoute.SubscriptionSettings,
+    subscriptions: List<SubscriptionSettingsUiState>,
+): SubscriptionSettingsUiState? {
+    val current = subscriptions.find { it.subId == route.subId }
+    var cached by remember(route.subId) { mutableStateOf(current) }
+    if (current != null) {
+        cached = current
+    }
+    return cached
```
inthewaves: The assignment during composition looks like it could be improved, since it's mutating Compose state during composition. I think something like this might be more pure and makes it more explicit that current is prioritized over cached:

```suggestion
    val current = subscriptions.find { it.subId == route.subId }
    var cached by remember(route.subId) { mutableStateOf(current) }
    SideEffect {
        if (current != null && cached != current) {
            cached = current
        }
    }
    return current ?: cached
```

app/src/androidTest/java/com/android/common/test/rules/MessagingTestRule.kt
```diff
@@ -0,0 +1,21 @@
...
+        description: Description,
+    ): Statement {
+        return object : Statement() {
+            override fun evaluate() {
+                TestDataSeeder.seedTestData()
+                base.evaluate()
+            }
+        }
+    }
+}
```
inthewaves, line 25: I think we should make sure seeded data is cleaned up after tests with `clearSeededTestData`

src/com/android/messaging/ui/appsettings/subscription/delegate/SubscriptionSettingsDelegate.kt
```diff
@@ -0,0 +1,132 @@
...
+                        trySend(Unit)
+                    }
+                }
+                manager.addOnSubscriptionsChangedListener(context.mainExecutor, listener)
+                awaitClose {
+                    manager.removeOnSubscriptionsChangedListener(listener)
+                }
+            }
+                .drop(1)
+                .collect { refresh() }
```
inthewaves: Using `callbackFlow` is nice here

For the `.drop(1)`, note that the Settings app in AOSP adds a caveat: https://cs.android.com/android/platform/superproject/+/android-latest-release:packages/apps/Settings/src/com/android/settings/network/telephony/SubscriptionRepository.kt;l=201-206

>    Note: Even the SubscriptionManager.addOnSubscriptionsChangedListener's doc says the
>    SubscriptionManager.OnSubscriptionsChangedListener.onSubscriptionsChanged() method will
>    also be invoked once initially when calling it, there still case that the
>    onSubscriptionsChanged() method is not invoked initially. For example, when the
>    onSubscriptionsChanged event never happens before, on a device never ever has any
>    subscriptions.

We should consider removing `.drop(1)` and set the other flow function used in AOSP settings app like `conflate` and adding `onAddListenerFailed()` to the listener

src/com/android/messaging/ui/appsettings/general/ui/AppSettingsScreen.kt
```diff
@@ -0,0 +1,168 @@
...
+                navigationIcon = {
+                    IconButton(onClick = onNavigateBack) {
+                        Icon(
+                            imageVector = Icons.AutoMirrored.Filled.ArrowBack,
+                            contentDescription = null,
+                        )
+                    }
+                },
+                scrollBehavior = scrollBehavior,
+            )
```
inthewaves: Perhaps this `LargeTopAppBar` could be refactored with the contentDescription set for the icon

src/com/android/messaging/ui/appsettings/subscription/ui/SubscriptionSettingsScreen.kt
```diff
@@ -0,0 +1,381 @@
...
+                navigationIcon = {
+                    IconButton(onClick = onNavigateBack) {
+                        Icon(
+                            imageVector = Icons.AutoMirrored.Filled.ArrowBack,
+                            contentDescription = null,
+                        )
+                    }
+                },
+                scrollBehavior = scrollBehavior,
+            )
```
inthewaves: Perhaps this `LargeTopAppBar` could be refactored with the contentDescription set for the icon

src/com/android/messaging/ui/appsettings/screen/SettingsMainScreen.kt
```diff
@@ -0,0 +1,87 @@
...
+                    IconButton(onClick = onNavigateBack) {
+                        Icon(
+                            imageVector = Icons.AutoMirrored.Filled.ArrowBack,
+                            contentDescription = null,
+                        )
+                    }
+                },
+                scrollBehavior = scrollBehavior,
+            )
+        },
```
inthewaves, line 40: Missing `contentDescription` for accessibility. [Documentation for androidx.compose.material3.Icon](https://developer.android.com/reference/kotlin/androidx/compose/material3/Icon.composable#Icon(androidx.compose.ui.graphics.vector.ImageVector,kotlin.String,androidx.compose.ui.Modifier,androidx.compose.ui.graphics.Color)) says

> 	contentDescription: text used by accessibility services to describe what this icon represents. This should always be provided unless this icon is used for decorative purposes, and does not represent a meaningful action that a user can take. This text should be localized, such as by using androidx.compose.ui.res.stringResource or similar

Since this is not for decorative purposes, we need to provide a description

Also, perhaps this `LargeTopAppBar` could be refactored with the contentDescription set for the icon

src/com/android/messaging/ui/appsettings/screen/SettingsEffectHandler.kt
```diff
@@ -0,0 +1,57 @@
...
+                }
+                context.startActivity(intent)
+            }
+
+            is SettingsScreenEffect.OpenNotificationSettings -> {
+                val intent = Intent(Settings.ACTION_APP_NOTIFICATION_SETTINGS).apply {
+                    putExtra(Settings.EXTRA_APP_PACKAGE, context.packageName)
+                }
+                context.startActivity(intent)
+            }
```
inthewaves: I think generally we should probably require the `Context` to be an `Activity` for the startActivity calls, especially for the RequestDefaultSmsApp requiring `startActivityForResult`. I don't think the startActivity works for e.g. an application context (without `FLAG_ACTIVITY_NEW_TASK` anyway), and `RequestDefaultSmsApp` requiring `startActivityForResult` needs `Activity`

src/com/android/messaging/ui/appsettings/ApplicationSettingsActivity.java
```diff
@@ -1,223 +0,0 @@
...
-        private boolean mIsSmsPreferenceClicked;
-
-        public ApplicationSettingsFragment() {
-            // Required empty constructor
-        }
-
-        @Override
-        public void onCreatePreferences(@Nullable Bundle savedInstanceState, String rootKey) {
-            getPreferenceManager().setSharedPreferencesName(BuglePrefs.SHARED_PREFERENCES_NAME);
-            addPreferencesFromResource(R.xml.preferences_application);
```
inthewaves, line 103: We can remove `res/xml/preferences_application.xml` now if possible

src/com/android/messaging/ui/appsettings/ApplicationSettingsActivity.java
```diff
@@ -1,223 +0,0 @@
...
-        ft.replace(android.R.id.content, new ApplicationSettingsFragment());
-        ft.commit();
-    }
-
-    @Override
-    public boolean onCreateOptionsMenu(Menu menu) {
-        if (super.onCreateOptionsMenu(menu)) {
-            return true;
-        }
-        getMenuInflater().inflate(R.menu.settings_menu, menu);
```
inthewaves, line 66: We can remove `res/menu/settings_menu.xml` now if possible

inthewaves (changes requested):

src/com/android/messaging/ui/appsettings/screen/model/SettingsUiState.kt
RankoR: +1, I'm not 100% sure, but I guess that Compose doesn't count `List` as Stable/Immutable, so it's better to use ImmutableList (I'm already using it on the Conversation screen). Replacing it is a simple mechanical refactor, so it's worth doing now.

inthewaves:
FYI: I can't resolve my own review comments thanks to a GitHub bug from 2021 (probably related to not having write access): https://github.com/orgs/community/discussions/8327

src/com/android/messaging/ui/appsettings/SettingsActivity.java
```diff
@@ -1,180 +0,0 @@
...
-        public void onCreate(final Bundle savedInstanceState) {
-            super.onCreate(savedInstanceState);
-            mBinding.bind(DataModel.get().createSettingsData(getActivity(), this));
-            mBinding.getData().init(LoaderManager.getInstance(this), mBinding);
-        }
-
-        @Override
-        public View onCreateView(final LayoutInflater inflater, final ViewGroup container,
-                final Bundle savedInstanceState) {
-            final View view = inflater.inflate(R.layout.settings_fragment, container, false);
```
inthewaves, line 100: could remove `res/layout/settings_fragment.xml` now too

src/com/android/messaging/ui/appsettings/screen/SettingsEffectHandler.kt
inthewaves: This requires changing Context to Activity to use `startActivityForResult`; see my later review comment at https://github.com/GrapheneOS/Messaging/pull/101#discussion_r3213732085

app/src/androidTest/java/com/android/messaging/ui/appsettings/screen/SettingsScreenTest.kt
```diff
@@ -0,0 +1,337 @@
...
+                defaultSmsAppLabel = "Messaging",
+                sendSoundEnabled = true,
+            ),
+            subscriptionSettings = listOf(
+                SubscriptionSettingsUiState(
+                    subId = 1,
+                    displayName = "Advanced Settings",
+                    displayDetail = "+1234567890",
+                ),
+            ),
```
inthewaves: Test compilation failure likely because these need to be updated to use immutable lists

inthewaves (changes requested):

inthewaves (approved):
Looks nice, seems to maintain parity with old Messaging app code as well

app/src/test/java/com/android/messaging/ui/appsettings/screen/SettingsViewModelTest.kt
```diff
@@ -0,0 +1,357 @@
...
+        subscriptionSettingsDelegate: SubscriptionSettingsDelegate =
+            FakeSubscriptionSettingsDelegate(),
+    ): SettingsViewModel {
+        return SettingsViewModel(
+            appSettingsDelegate = appSettingsDelegate,
+            subscriptionSettingsDelegate = subscriptionSettingsDelegate,
+        )
+    }
+
+    private class FakeAppSettingsDelegate : AppSettingsDelegate {
```
RankoR: `Fake`s should be replaced with mocks, probably.

src/com/android/messaging/ui/appsettings/general/mapper/AppSettingsUiStateMapper.kt
```diff
@@ -0,0 +1,49 @@
...
+import android.content.Context
+import com.android.messaging.R
+import com.android.messaging.ui.appsettings.general.model.AppSettingsUiState
+import com.android.messaging.util.BuglePrefs
+import com.android.messaging.util.DebugUtils
+import com.android.messaging.util.PhoneUtils
+import dagger.hilt.android.qualifiers.ApplicationContext
+import javax.inject.Inject
+
+internal interface AppSettingsUiStateMapper {
```
RankoR, line 10: Mapper converts one entity type to another; This class looks more like a kind of repository that creates a UI-layer object.
It would be cleaner if the prefs retrieval were done on the data layer and then transformed into the UI state on some upper level.

src/com/android/messaging/ui/appsettings/screen/SettingsNavRouteSavedState.kt
```diff
@@ -0,0 +1,54 @@
...
+    }
+}
+
+private fun SettingsNavRoute.toSavedState(): SettingsNavRouteSavedState {
+    return when (this) {
+        SettingsNavRoute.Main -> SettingsNavRouteSavedState.Main
+
+        SettingsNavRoute.AppSettings -> SettingsNavRouteSavedState.AppSettings
+
+        is SettingsNavRoute.SubscriptionSettings -> SettingsNavRouteSavedState.SubscriptionSettings(
```
RankoR: Please wrap multiline `when` branches with `{}`:

```
is SettingsNavRoute.SubscriptionSettings -> {
   SettingsNavRouteSavedState.SubscriptionSettings(...
}
```

(in other cases too)

src/com/android/messaging/ui/appsettings/screen/SettingsScreen.kt
```diff
@@ -0,0 +1,255 @@
...
+import com.android.messaging.ui.appsettings.subscription.model.SubscriptionSettingsUiState
+import com.android.messaging.ui.appsettings.subscription.ui.SubscriptionSettingsScreen
+import kotlinx.collections.immutable.ImmutableList
+
+private const val SLIDE_OFFSET_DIVISOR = 3
+
+@Composable
+internal fun SettingsScreen(
+    effectHandler: SettingsEffectHandler,
+    onNavigateBack: (() -> Unit),
```
RankoR: Redundant outer parentheses

RankoR (changes requested):
Haven't checked the parameters' stability yet.

RankoR (approved):

PR #109, opened 2026-05-01
Rewrite Conversation screen with Kotlin and Compose

RankoR:
First (of 3) PR with Conversation screen rewritten with Kotlin and Compose.

I've split it into 3 PRs:

- This one, containing the main code changes
- Compose `@Preview`s
- Tests

This split is required because it's a very large rewrite, so it makes it easier to review. I will create the remaining two after this one is merged.

For reference, all changes together are in https://github.com/RankoR-GOS/Messaging/commits/task/coversation-compose.

Here's a demo (the lags are coming from my emulator, not from the app itself):

https://github.com/user-attachments/assets/a6b12717-6080-44df-92d3-859f2f91dc6c


inthewaves (changes requested):
Looks good so far, haven't looked too deeply at the code yet. Some things from testing it:
1. One bug I've noticed is that if a message contains a hyperlink, selecting the message makes the hyperlink invisible (videos below show this)
2. I think we should hide the number/name on the incoming message bubbles if the conversation is a 1-1 thread
3. Message details dialog popup looks untouched (e.g. doesn't follow dark theme)
4. I didn't intentionally set it up this way, but my colors for Material You under OS Settings > Wallpaper & style > Colors were setup like this:
    Image

   With these colors, the contrast for selected outgoing messages is hard to see in dark mode. Also link color contrast issue makes links hard to see (link is https://example.com): 

    https://github.com/user-attachments/assets/978ea077-3728-401b-b7a9-f6d707a4bdcf

5.  I can't long press a message containing just a link. Generally, long presses don't select when the pointer is over a link. I think ideally long pressing a link should just select the message instead of opening the link as well; this matches the old Messages behavior

    https://github.com/user-attachments/assets/ca8794ea-6714-4675-a149-613c5a844ca5

6. Showing the phone number in the top app bar is debatable as the old Messages app doesn't do it. Google Messages and Signal also hides the number in the app bar. My preference is that we don't show it

    As a side note, Signal used to also show the number at the top, and this also used to be a talking point for many users e.g. https://community.signalusers.org/t/option-to-hide-contact-phone-number-on-the-top-of-the-chat/4455/7. Signal finally hid the number at the top in https://community.signalusers.org/t/option-to-hide-contact-phone-number-on-the-top-of-the-chat/4455/38

thomasbuilds:
I got interested in the PR and found a bug:

`ConversationDraftStore.getSelfParticipantId()` declares a nullable return, and every caller treats `null` as "no self available, bail out". But the implementation maps `null` → `""`:

```kotlin
return conversation.selfId.orEmpty()
```

`ConversationListItemData.selfId` is genuinely null in real flows (newly‑created threads before any send, multi‑SIM before a SIM is picked, after SIM swap/removal). The empty string then defeats the null‑guards in `ConversationDraftsRepository`:

- **Load** (`loadConversationDraft`, L112–L115): `?: return ConversationDraft()` never fires; the DB is then queried with `WHERE self_id = ''`, returns nothing, and any draft saved under the real `self_id` appears to vanish until `current_self_id` is populated.
- **Save** (`bindDraftParticipantsIfNeeded` → `bindMissingDraftParticipants`, L215–L250): the mapper correctly produces `message.selfId = null` for a blank draft self id, but `bindMissingDraftParticipants` then overwrites it with `""`, persisting `self_id = ""` and `participant_id = ""` into the draft message row. That row will never match participant lookups on the next load.

**Suggested fix**

Honor the `String?` contract:

```suggestion
        return conversation.selfId?.takeIf { it.isNotBlank() }
```

Optional defence‑in‑depth in `ConversationDraftsRepository.bindMissingDraftParticipants`:

```kotlin
if (selfParticipantId.isBlank()) return message
```

RankoR:
@inthewaves fixed everything. 

> With these colors, the contrast for selected outgoing messages is hard to see in dark mode

There's no good universal solution with color-only selection highlighting. Added a "checkbox" selection indicator.

> Message details dialog popup looks untouched

This is intentional, same as the "conversation details" screen. This PR is way too large, even without it. The rewrite includes only the parts that are required and can't be left as-is without hacks like including legacy views in the Compose code.

@thomasbuilds good catch, thanks, fixed.

src/com/android/messaging/ui/conversation/ConversationFragment.java
```diff
@@ -1,1661 +0,0 @@
...
-                @Override
-                public void onClick(final DialogInterface dialog, final int which) {
-                    showAttachmentChooser(conversationId, activity);
-                }});
-        } else {
-            builder.setMessage(R.string.attachment_limit_reached_dialog_message_when_composing)
-                    .setPositiveButton(android.R.string.ok, null);
-        }
-        builder.show();
-    }
```
inthewaves, line 1523: I think we should reimplement these attachment count and total size validation checks somehow (at the very least, a `rg`/`grep` of these strings shows that these strings are no longer used)

src/com/android/messaging/ui/conversation/ComposeMessageView.java
```diff
@@ -1,1016 +0,0 @@
...
-                if (mHost.shouldShowSubjectEditor()) {
-                    showSubjectEditor();
-                } else {
-                    boolean shown = mInputManager.toggleSimSelector(true /* animate */,
-                            getSelfSubscriptionListEntry());
-                    hideAttachmentsWhenShowingSims(shown);
-                }
-                return true;
-            }
-        });
```
inthewaves, line 241: In old Messages screen, long pressing send button has these options
- Long pressing brings up a Subject menu which also seems to turn the message into MMS. 

  This appears to be omitted in the rewrite; perhaps we should bring it back in some form. Google Messages puts it in the top app bar menu under "Show subject field" that brings a dialog for it

    Image

- For multi-SIM, it shows the SIM selector. 

  The rewrite already moves this to the app bar in top-right menu

src/com/android/messaging/ui/conversation/metadata/ui/ConversationTopAppBar.kt
```diff
@@ -0,0 +1,606 @@
...
+    onSimSelectorClick: () -> Unit,
+    onItemClick: (() -> Unit) -> Unit,
+) {
+    ConversationTopAppBarOverflowMenuItem(
+        isVisible = visibility.isSimSelectorVisible,
+        testTag = CONVERSATION_SIM_SELECTOR_MENU_ITEM_TEST_TAG,
+        label = simSelectorLabel,
+        icon = Icons.Rounded.SimCard,
+        onClick = { onItemClick(onSimSelectorClick) },
+    )
```
inthewaves, line 360: For SIM selector, I think it's a bit inconsistent with the other menu options. The other menu options have text like "Add people", "Archive", "Delete", so these are all verbs. The SIM just displays the SIM name, so it might not be immediately clear that it's for switching SIM. Perhaps if the overflow menu supports a subtitle of some sort

Note that Google Messages for reference puts it on the conversation details screen, and make it clear that switch is the action

Image

src/com/android/messaging/ui/conversation/messages/ui/ConversationMessages.kt
```diff
@@ -0,0 +1,346 @@
...
+                ),
+                isSelectionMode = selectedMessageIds.isNotEmpty(),
+                isSelected = selectedMessageIds.contains(message.messageId),
+                showIncomingSenderLabels = showIncomingSenderLabels,
+                onAttachmentClick = onAttachmentClick,
+                onExternalUriClick = onExternalUriClick,
+                onMessageClick = onMessageClick,
+                onMessageLongClick = onMessageLongClick,
+                onMessageResendClick = onMessageResendClick,
+            )
```
inthewaves: I think we should distinguish messages sent by different SIM more. Currently it looks like this

Image

For example, in Google Messages (the SIM name itself is a link that opens the conversation settings where the SIM selector is)

Image

In old messages app (note that it also puts the "Send message from SIM1" below, but I think Google Message design makes that more explicit, since it acts like a divider between messages from SIM1 and messages from SIM2

Image

src/com/android/messaging/data/conversation/repository/ConversationsRepository.kt
```diff
@@ -0,0 +1,363 @@
...
+
+    override fun deleteConversation(conversationId: String) {
+        if (conversationId.isBlank()) {
+            return
+        }
+
+        DeleteConversationAction.deleteConversation(
+            conversationId,
+            System.currentTimeMillis(),
+        )
```
inthewaves, line 206: `System.currentTimeMillis()` is used for deletion cutoff unconditionally, but old code in `src/com/android/messaging/datamodel/data/ConversationData.java` says "If possible use timestamp of last message shown to delete only messages user is aware of "

https://github.com/GrapheneOS/Messaging/blob/caf27084c6bed3df95eb664f4fdad095fd11b828/src/com/android/messaging/datamodel/data/ConversationData.java#L630-L639

`mConversationMetadata.deleteConversation()` path uses `DeleteConversationAction.deleteConversation(mConversationId, mTimestamp);`

src/com/android/messaging/ui/conversation/ConversationFragment.java
```diff
@@ -1,1661 +0,0 @@
...
-    public void deleteConversation() {
-        if (isReadyForAction()) {
-            final Context context = getActivity();
-            mBinding.getData().deleteConversation(mBinding);
-            closeConversation(mConversationId);
-        } else {
-            warnOfMissingActionConditions(false /*sending*/,
-                    null /*commandToRunAfterActionConditionResolved*/);
-        }
-    }
```
inthewaves, line 1178: we should make sure we have similar `isReadyForAction()` checks in new code

src/com/android/messaging/ui/conversation/ComposeMessageView.java
```diff
@@ -1,1016 +0,0 @@
...
-                    hideSimSelector();
-                }
-            }
-        });
-
-        // onFinishInflate() is called before self is loaded from db. We set the default text
-        // limit here, and apply the real limit later in updateOnSelfSubscriptionChange().
-        mComposeEditText.setFilters(new InputFilter[] {
-                new LengthFilter(MmsConfig.get(ParticipantData.DEFAULT_SELF_SUB_ID)
-                        .getMaxTextLimit()) });
```
inthewaves, line 218: Old code has SMS text limits and splits, and it also shows UI when conversation reaches limit to tell when it's going to split. We should probably add these UI elements back

The `112 / 2` above the send button is what is meant here; this is from Google Messages:

Image

src/com/android/messaging/ui/conversation/ComposeMessageView.java
```diff
@@ -1,1016 +0,0 @@
...
-
-    private void updateOnSelfSubscriptionChange() {
-        // Refresh the length filters according to the selected self's MmsConfig.
-        mComposeEditText.setFilters(new InputFilter[] {
-                new LengthFilter(MmsConfig.get(mBinding.getData().getSelfSubId())
-                        .getMaxTextLimit()) });
-        mComposeSubjectText.setFilters(new InputFilter[] {
-                new LengthFilter(MmsConfig.get(mBinding.getData().getSelfSubId())
-                        .getMaxSubjectLength())});
-    }
```
inthewaves, line 532: Old code has SMS text limits and splits, and it also shows UI when conversation reaches limit to tell when it's going to split. We should probably add these UI elements back

src/com/android/messaging/ui/conversation/recipientpicker/component/RecipientSelectionContactsContent.kt
```diff
@@ -0,0 +1,342 @@
...
+            }
+        }
+
+        recipientSelectionContactItems(
+            uiState = uiState,
+            rowDecorators = rowDecorators,
+            onRecipientClick = onRecipientClick,
+            onRecipientLongClick = onRecipientLongClick,
+        )
+    }
```
inthewaves, line 149: Two things:
- When a recipient has multiple numbers, I think we should combine them into the same contact. 
- It looks like the numbers are not normalized from what the user input into the Contacts app

Currently it looks like this:

Image

Note that the numbers have different formatting corresponding to how those numbers were entered into the contact app. At least for North American numbers (hence the `+1`), I entered the format as `+1xxxxxxxxxx`, `+1 xxx-xxx-xxxx`, `(xxx) xxx-xxxx` in the Contacts app

This is what old messages screen does; all numbers are clickable and normalized, and labelled with what the use put for those numbers in Contacts

Image

Google Messages is also normalized, and multiple contacts are put into that +x more part, although selecting one of them looks a bit weird in Google Messages since it doesn't display the actual number that gets selected

Image

Also a weird edge case but I think this is a preexisting bug since it also happens in old messages: When you have two contacts with same number but you format them differently in contacts, only one of them gets selected if you long press in new chat screen. But if you format them the same, both of them get selected (Google Messages behavior)

Another edge case with contact linking in the Contacts app: I noticed this because somehow my phone ended up with two contacts linked together with same number, one with no name and one with it. When both of them were linked in the Contacts app, both old Messages app and Google Messages groups them under the same contact. When they're unlinked, they're shown as separate rows, even if they have the same number. The number is display twice in old Messages app, but Google Messages managed to dedupe it (see the pictures above; the `+1xxxxxxxxxx` and `+1 xxx-xxx-xxxx` numbers are the same but just formatted differenlt).

src/com/android/messaging/ui/conversation/recipientpicker/RecipientSelectionContent.kt
```diff
@@ -0,0 +1,119 @@
...
+
+            RecipientSelectionQueryField(
+                query = uiState.picker.query,
+                enabled = uiState.isQueryEnabled,
+                prefixText = strings.queryPrefixText,
+                placeholderText = strings.queryPlaceholderText,
+                onQueryChanged = onQueryChanged,
+            )
+
+            Spacer(modifier = Modifier.height(height = 12.dp))
```
inthewaves: For multi-SIM setups, it might be good to add a way to choose the SIM here in new chat screen like Google Messages

Image

src/com/android/messaging/ui/conversation/ComposeMessageView.java
```diff
@@ -1,1016 +0,0 @@
...
-                                playSentSound();
-                                mHost.sendMessage(message);
-                                hideSubjectEditor();
-                                if (AccessibilityUtil.isTouchExplorationEnabled(getContext())) {
-                                    AccessibilityUtil.announceForAccessibilityCompat(
-                                            ComposeMessageView.this, null,
-                                            R.string.sending_message);
-                                }
-                            }
-                            break;
```
inthewaves, line 411: Could be good to incorporate sound and accessibility announcement again

src/com/android/messaging/ui/conversation/messages/ui/message/ConversationMessage.kt
```diff
@@ -0,0 +1,377 @@
...
+    return when (status) {
+        Status.Outgoing.Delivered -> R.string.delivered_status_content_description
+        Status.Outgoing.Sending -> R.string.message_status_sending
+        Status.Outgoing.Resending -> R.string.message_status_send_retrying
+        Status.Outgoing.AwaitingRetry -> R.string.message_status_failed
+        Status.Outgoing.Failed -> R.string.message_status_send_failed
+        Status.Outgoing.FailedEmergencyNumber -> {
+            R.string.message_status_send_failed_emergency_number
+        }
+        Status.Incoming.YetToManualDownload -> R.string.message_status_download
```
inthewaves: I think we need to show more info when MMS needs to be downloaded. It should show the same amount of info as the old messages app

Old messages app:

Image

New:

Image

Also 
- Tapping it to download doesn't work. Download option is only in the app bar
- The UI seems to merge the two MMS download status for the last two messages, but only the bottom one is downloading. The top one should be in tap to download state like old message app

Google Messages for reference (I think this doesn't look ideal and actually looks incomplete; old messages app style is the best IMO):

Image

src/com/android/messaging/ui/conversation/mediapicker/ConversationMediaPicker.kt
```diff
@@ -0,0 +1,363 @@
...
+        sheetState = sheetState,
+        embeddedPhotoPickerState = embeddedPhotoPickerState,
+    )
+
+    ConversationMediaPickerScaffoldContent(
+        modifier = modifier,
+        cameraController = cameraController,
+        scaffoldState = scaffoldState,
+        embeddedPhotoPickerState = embeddedPhotoPickerState,
+        embeddedPhotoPickerFeatureInfo = rememberConversationEmbeddedPhotoPickerFeatureInfo(),
```
inthewaves, line 280: Android Studio flags `rememberConversationEmbeddedPhotoPickerFeatureInfo` since it has `@RequiresExtension(extension = Build.VERSION_CODES.UPSIDE_DOWN_CAKE, version = 15)`:

> Call requires version 15 of the U Extensions SDK (current min is 0):

minSdk 35 covers the eligibility but doesn't guarantee it https://developer.android.com/guide/sdk-extensions

Also check `EmbeddedPhotoPicker`

src/com/android/messaging/ui/conversation/composer/ui/ConversationComposeMessageField.kt
```diff
@@ -0,0 +1,320 @@
...
+
+@Composable
+private fun ConversationComposeAttachmentMenu(
+    modifier: Modifier = Modifier,
+    enabled: Boolean,
+    isAudioRecordActionEnabled: Boolean,
+    onContactAttachClick: () -> Unit,
+    onMediaPickerClick: () -> Unit,
+    onAudioAttachClick: () -> Unit,
+) {
```
inthewaves, line 200: Menu looks like it's too high

Image

inthewaves (changes requested):

inthewaves (changes requested):
Also, currently we're not showing any profile pictures in the chat list. We should show profile pictures for other members in MMS groups; can keep the profile picture out of the chat list for 1-1 conversations. Old messages app showed profile pictures all the time in the conversations list which I think is too much.

Signal and Google Messages for reference:

Image

Image

src/com/android/messaging/ui/conversation/messages/ui/ConversationMessages.kt
inthewaves: Actually I think going with the design of the old Messages screen is easier and makes more sense for responses:

image

Google Messages divider-like UI seems unclear when responding to each SIM's number. It seems to switch to the SIM that's receiving the message only after it arrives:

image

src/com/android/messaging/ui/conversation/metadata/ui/ConversationTopAppBar.kt
RankoR: >  The SIM just displays the SIM name, so it might not be immediately clear that it's for switching SIM

I agree. What wording would you suggest for the menu item?

> Note that Google Messages for reference puts it on the conversation details screen, and make it clear that switch is the action

It's a good solution, but I would not change the conversation details screen in this iteration.

inthewaves: Maybe `Switch from $SIM`, although I'm not sure how that does with long SIM names.

Or if subtitles are supported with that component, maybe something like

> Switch SIMs
> $SIMNAME (in slighty smaller font)

src/com/android/messaging/ui/conversation/ComposeMessageView.java
RankoR: Added support for subjects, like it's done in Google Messages, not with a long press (without message priorities for now).

image

Also added multi-sim selector on long press.

src/com/android/messaging/ui/conversation/ConversationFragment.java
RankoR: Done. Works as a hard limit, checking before and after adding any attachment.

src/com/android/messaging/ui/conversation/metadata/ui/ConversationTopAppBar.kt
RankoR: Done.

inthewaves:
FYI: I can't resolve my own review comments thanks to a GitHub bug from 2021 (probably related to not having write access): https://github.com/orgs/community/discussions/8327

src/com/android/messaging/ui/conversation/messages/ui/ConversationMessages.kt
RankoR: Fixed

image

src/com/android/messaging/ui/conversation/composer/ui/ConversationSendActionButtonGesture.kt
```diff
@@ -48,29 +54,56 @@ internal fun Modifier.conversationSendActionButtonGesture(
...
-                )
+            val isLockedRecording = currentIsRecordingActive && currentIsRecordingLocked
+
+            when {
+                mode == ConversationSendActionButtonMode.Send -> {
+                    handleSendModeLongPress(
+                        hapticFeedback = hapticFeedback,
+                        onSendActionLongClick = currentOnSendActionLongClick,
+                    )
+                }
```
inthewaves: There appears to be a bug (not sure on location) where long pressing the send button to show the bottom sheet for SIM selection will both show the bottom sheet and send the message with current SIM. I think it should only send when user presses the send button normally; long pressing should not send message

inthewaves (changes requested):

src/com/android/messaging/ui/conversation/composer/ui/ConversationSendActionButtonGesture.kt
RankoR: You're right, there's a bug.

src/com/android/messaging/data/conversation/repository/ConversationsRepository.kt
RankoR: Fixed

src/com/android/messaging/ui/conversation/ConversationFragment.java
RankoR: Same checks were already performed on sending/retrying; added them to the deleting action.

src/com/android/messaging/ui/conversation/ComposeMessageView.java
RankoR: Done

src/com/android/messaging/ui/conversation/recipientpicker/component/RecipientSelectionContactsContent.kt
RankoR: Contacts with multiple destinations are now grouped; Phone numbers are normalized and then formatted.

m4pl:
I haven't reviewed the code itself yet, but I found several bugs during UI testing:

**Emulator settings:**
Screenshot 2026-05-12 at 15 34 06

1. **Message jumping:** When selecting messages with a small font size, the messages jump around instead of staying in position.
[Screen_recording_20260512_153503.webm](https://github.com/user-attachments/assets/a63cf15c-72ca-4a23-ab59-16372b67d064)

2. **Keyboard interaction during recording:** While recording an audio message, user can still tap the input field, which opens the keyboard. I believe this action should be blocked during recording
[Screen_recording_20260512_154245.webm](https://github.com/user-attachments/assets/5fdd25e5-311a-4fc6-9da2-1770f071b37c)

3. **SIM switch:** Long-pressing the "Send" icon automatically sends the message, even if the user only intended to switch the SIM card.
[Screen_recording_20260512_165927.webm](https://github.com/user-attachments/assets/517bab46-ac83-4fcc-8b59-bb4918f82e6c)

4. **Contact issues:** The avatar for attached contacts is not appearing. Also, if a contact fails to send (which happens frequently in my tests), clicking the message attempts to resend it, but the label still says "View contact card."
[Screen_recording_20260512_172857.webm](https://github.com/user-attachments/assets/d97134d7-819d-4d6b-8c0b-da19ecaa39b8)

---

Minor UI Suggestions:

1. **Toolbar height:** The contact selection area at the top looks too small. It should fill the full height of the toolbar as it did before.
Screenshot_20260512_153628

2. **Attachments spacing:** There is too much vertical space between the attachments popup and the message input.
Screenshot_20260512_154425

3. **Button visibility:** Media selector buttons are dark and hard to read, especially when the camera is focused on a dark area.
Screenshot_20260512_154841

4. **Input padding:** The message input field has excessive top and bottom padding.
Screenshot_20260512_160222

5. **Contact attachments spacing:** The spacing between attached contacts within a single message is too large and should be reduced.
Screenshot_20260512_161709

6. **Selection marker:** In my opinion, the left padding for the "selected message" marker is a bit too wide.
Screenshot_20260512_163504

7. **Icon quality:** The "Stop Recording" button appears to be low-resolution.
Screenshot_20260512_154313

src/com/android/messaging/ui/conversation/recipientpicker/RecipientSelectionContent.kt
RankoR: Done

src/com/android/messaging/ui/conversation/ComposeMessageView.java
RankoR: Done

src/com/android/messaging/ui/conversation/messages/ui/message/ConversationMessage.kt
RankoR: image

src/com/android/messaging/ui/conversation/mediapicker/ConversationMediaPicker.kt
inthewaves: Further context for others: In AP3A (prior to Dec 2024), SDK extension version was lower than 15 and would require a Play system update to get the new extension version for embedded photo picker support: https://github.com/GrapheneOS/platform_build_release/blob/16-qpr2/flag_values/ap3a/RELEASE_PLATFORM_SDK_EXTENSION_VERSION.textproto

Not an issue for AP4A (Dec 2024) and onwards including recent GrapheneOS builds: https://github.com/GrapheneOS/platform_build_release/blob/16-qpr2/flag_values/ap4a/RELEASE_PLATFORM_SDK_EXTENSION_VERSION.textproto

logic in `packages/modules/SdkExtensions` reads from `/apex/*/etc/sdkinfo.pb` (appears to be from `PLATFORM_SDK_EXTENSION_VERSION`, e.g. see `build/make/core/version_util.mk:93` which is used in `packages/modules/common/tools/Android.bp` to run `packages/modules/common/tools/gen_sdkinfo.py`)

`packages/modules/SdkExtensions` also reads from `/apex/com.android.sdkext/etc/extensions_db.pb` (this seems it would get updated in a Google Play system update along with the APEX for `com.android.mediaprovider` for embedded photo picker), and `packages/modules/SdkExtensions/gen_sdk/extensions_db.textpb`

src/com/android/messaging/ui/conversation/composer/ui/ConversationSendActionButtonGesture.kt
RankoR: Fixed

src/com/android/messaging/ui/conversation/composer/ui/ConversationComposeMessageField.kt
RankoR: Fixed

src/com/android/messaging/ui/conversation/mediapicker/ConversationMediaPicker.kt
RankoR: Agreed to ignore it, as it doesn't affect GrapheneOS users who will use this app.

m4pl:
I also noticed several UI issues when microphone and camera permissions are revoked:

**Media selector:** The dialog requesting camera permission is layered behind the UI elements.
Screenshot_20260513_235758

**Audio recording:** If permission is granted during the recording process, the recording behavior breaks, making it difficult to cancel the action.
[Screen_recording_20260513_234755.webm](https://github.com/user-attachments/assets/8a33bda5-c4f9-4dcb-817c-85f15f25c7c8) (emulator is lagging a bit)

RankoR:
Everything fixed, except 

> Input padding: The message input field has excessive top and bottom padding.

It's visually ~ the same as in Google Messages (maybe a bit larger), doesn't look bad, especially on Pixel-size screens, but reducing the padding seems to require quite complex changes compared to the outcome.

src/com/android/messaging/ui/conversation/recipientpicker/component/simselector/NewChatSimSelector.kt
```diff
@@ -0,0 +1,231 @@
...
+                        color = MaterialTheme.colorScheme.onSurface,
+                    )
+
+                    subscription.displayDestination?.let { destination ->
+                        Text(
+                            text = destination,
+                            style = MaterialTheme.typography.bodyMedium,
+                            color = MaterialTheme.colorScheme.onSurfaceVariant,
+                        )
+                    }
```
inthewaves: Appears to be squished:

Image

src/com/android/messaging/ui/conversation/recipientpicker/component/RecipientSelectionContent.kt
```diff
@@ -0,0 +1,166 @@
...
+    ) {
+        Column(modifier = Modifier.fillMaxWidth()) {
+            RecipientSelectionQueryField(
+                query = query,
+                enabled = enabled,
+                prefixText = prefixText,
+                placeholderText = placeholderText,
+                onQueryChanged = onQueryChanged,
+                focusRequester = focusRequester,
+            )
```
inthewaves: Currently we don't have any indication of which contacts are selected in the new chat screen other than the checkmarks / selected background on the contacts in the list. Selected contacts can get hidden when e.g. one tries to search another contact and the results will hide the previously select contacts

Old Messaging app has the selected contacts directly in the search query field, and each one can be pressed to remove from a dialog that pops up

Image

Google Messages also has the same with the query hint changing to "Anyone else?"; tapping one to remove animates the pill/chip(?) to show the X button for removal instead of a popup

Image

src/com/android/messaging/ui/conversation/recipientpicker/component/RecipientSelectionContactAvatar.kt
```diff
@@ -0,0 +1,220 @@
...
+import com.android.messaging.ui.conversation.recipientpicker.model.RecipientPickerListItem
+
+@Composable
+internal fun RecipientSelectionContactAvatar(
+    item: RecipientPickerListItem,
+    isSelected: Boolean,
+) {
+    val avatarScale by rememberRecipientSelectionContactAvatarScale(
+        isSelected = isSelected,
+    )
```
inthewaves, line 44: For contacts with multiple numbers, I feel like we could also show a (smaller) checkmark as well so that it's more consistent with single-number contact selections

Image

inthewaves (changes requested):

src/com/android/messaging/ui/conversation/recipientpicker/component/RecipientSelectionContent.kt
```diff
@@ -0,0 +1,166 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.FocusRequester
+import androidx.compose.ui.focus.focusRequester
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.dp
+
+private val searchCardShape = RoundedCornerShape(size = 22.dp)
```
inthewaves: Maybe we can replace this with something in `MaterialTheme.shapes`, unless the dps there aren't really usable

This generally applies to all `RoundedCornerShape` usages

https://github.com/material-components/material-components-android/blob/master/docs/theming/Shape.md

https://developer.android.com/reference/kotlin/androidx/compose/material3/Shapes

https://developer.android.com/codelabs/jetpack-compose-theming#6 has examples on defining custom shapes

inthewaves (changes requested):

src/com/android/messaging/data/conversation/repository/ConversationsRepository.kt
```diff
@@ -0,0 +1,365 @@
...
+
+    fun unarchiveConversation(conversationId: String)
+
+    fun deleteConversation(conversationId: String, cutoffTimestamp: Long)
+}
+
+internal class ConversationsRepositoryImpl @Inject constructor(
+    private val contentResolver: ContentResolver,
+    @param:IoDispatcher
+    private val ioDispatcher: CoroutineDispatcher,
```
inthewaves: We have to be careful when using Kotlin `Dispatchers.IO` with SQLite. `MessagingContentProvider` has SQLite work. 

- https://blog.p-y.wtf/parallelism-with-android-sqlite covers  why SQLite work on `Dispatchers.IO` can occupy many IO threads while Android SQLite is still limited by its connection pool.
- https://medium.com/androiddevelopers/threading-models-in-coroutines-and-android-sqlite-api-6cab11f7eb90 explains the related transaction/thread-confinement problem. Coroutine dispatchers do not by themselves make Android SQLite access safe or parallel. It goes over the design decisions of how the androidx Room library handles Kotlin coroutines and SQLite.

For this code, I’d prefer something like `@MessagingDbDispatcher` / `@DatabaseDispatcher`, capped to the database concurrency: probably `1` (`.limitedParallelism(1)` or a dedicated Executor for the coroutine context) unless WAL is explicitly enabled, or the WAL pool size if it is. Generic file/media IO can stay on `@IoDispatcher`.

src/com/android/messaging/ui/conversation/attachment/ui/ConversationMediaThumbnailBitmapLoader.kt
```diff
@@ -0,0 +1,306 @@
...
+private const val SECOND_PASS_MINIMUM_SIZE = 24
+
+internal suspend fun loadConversationMediaThumbnailBitmap(
+    contentResolver: ContentResolver,
+    contentUri: Uri,
+    contentType: String,
+    size: IntSize,
+    softenBitmap: Boolean,
+): Bitmap? {
+    return withContext(context = Dispatchers.IO) {
```
inthewaves, line 30: There's some places where this seems to use `Dispatchers.IO` directly instead of from the DI-injected `@IoDispatcher`

inthewaves (changes requested):

src/com/android/messaging/ui/conversation/recipientpicker/component/simselector/NewChatSimSelector.kt
RankoR: Fixed

src/com/android/messaging/ui/conversation/recipientpicker/component/RecipientSelectionContent.kt
RankoR: https://github.com/user-attachments/assets/ee296fff-4615-4665-8e74-cebe2f7ea91e

src/com/android/messaging/ui/conversation/recipientpicker/component/RecipientSelectionContactAvatar.kt
RankoR: Fixed (see above)

src/com/android/messaging/ui/conversation/attachment/ui/ConversationMediaThumbnailBitmapLoader.kt
RankoR: It's an intentional trade-off; it's not ideal, but I don't see much sense in doing it some other way, at least for now.

src/com/android/messaging/data/conversation/repository/ConversationsRepository.kt
RankoR: Fixed

src/com/android/messaging/ui/conversation/recipientpicker/component/RecipientSelectionContent.kt
RankoR: Changed to the default shapes where possible. No need to change the default ones.

src/com/android/messaging/ui/conversation/composer/mapper/ConversationComposerUiStateMapper.kt
```diff
@@ -0,0 +1,144 @@
...
+    ): ConversationSegmentCounterUiState? {
+        val isSms = sendProtocol == ConversationDraftSendProtocol.SMS
+        val messageText = draft.messageText
+
+        if (!isSms || messageText.isBlank()) {
+            return null
+        }
+
+        val stats = MessageTextStats().apply {
+            updateMessageTextStats(ParticipantData.DEFAULT_SELF_SUB_ID, messageText)
```
inthewaves: For multi-SIM correctness, think this should use the relevant selected subscription when available instead of default sub id

Old code uses `getSelfSubId()`: 

https://github.com/GrapheneOS/Messaging/blob/f1e0efc2443d195380f5b2d656bb94d0151f93df/src/com/android/messaging/datamodel/data/DraftMessageData.java#L248-L255

src/com/android/messaging/ui/conversation/messages/delegate/ConversationMessageSelectionDelegate.kt
```diff
@@ -0,0 +1,563 @@
...
+
+    override fun onMessageClick(messageId: String) {
+        if (state.value.isSelectionMode) {
+            toggleMessageSelection(messageId = messageId)
+        }
+    }
+
+    override fun onMessageDownloadClick(messageId: String) {
+        conversationsRepository.downloadMessage(messageId = messageId)
+    }
```
inthewaves, line 111: Should extend action requirements to message downloads

Old code ref guards with `isReadyToSend`:

https://github.com/GrapheneOS/Messaging/blob/f1e0efc2443d195380f5b2d656bb94d0151f93df/src/com/android/messaging/ui/conversation/ConversationFragment.java#L1282-L1285

src/com/android/messaging/ui/conversation/messages/delegate/ConversationMessageSelectionDelegate.kt
```diff
@@ -0,0 +1,563 @@
...
+
+        clearMessageSelection()
+    }
+
+    private fun downloadSelectedMessage() {
+        val selectedMessage = singleSelectedMessageOrNull() ?: return
+
+        clearMessageSelection()
+        conversationsRepository.downloadMessage(selectedMessage.messageId)
+    }
```
inthewaves, line 238: Should extend action requirements to message downloads

src/com/android/messaging/ui/conversation/messages/delegate/ConversationMessageSelectionDelegate.kt
```diff
@@ -0,0 +1,563 @@
...
+        if (selectedMessageIds.isEmpty()) {
+            return
+        }
+
+        messageSelectionState.update { currentState ->
+            currentState.copy(
+                pendingDeleteMessageIds = selectedMessageIds,
+            )
+        }
+    }
```
inthewaves, line 295: should extend action requirements to message deletion (I can still delete messages in the app when not default SMS role)

Old code ref with `isReadyForAction`:

https://github.com/GrapheneOS/Messaging/blob/f1e0efc2443d195380f5b2d656bb94d0151f93df/src/com/android/messaging/ui/conversation/ConversationFragment.java#L1146-L1167

src/com/android/messaging/ui/conversation/screen/ConversationScreenEffects.kt
```diff
@@ -0,0 +1,346 @@
...
+                view = view,
+                snackbarHostState = snackbarHostState,
+                hostBoundsState = hostBoundsState,
+                effect = effect,
+                launchRoleRequest = defaultSmsRoleLauncher::launch,
+                onNavigateBack = onNavigateBack,
+                onDraftSent = { draftSentTick.intValue++ },
+            )
+        }
+    }
```
inthewaves: It seems like this `LauchedEffect` essentially only depends on `screenModel`, but it's keyed by all the other state that is strictly not related to the actual `Flow` being collected. The keys other than `screenModel` are just used in the `Flow` collection

If one of those keys changes, especially the `onNavigateBack` lambda, Compose cancels the current collector and starts a new one. Technically, since `screenModel.effects` is a `MutableSharedFlow(extraBufferCapacity = 1)` with the default `replay = 0`, effects emitted while the collector is cancelled/resubscribing are not replayed and can be missed. Cancellation can also interrupt a suspend handler already in progress

We can use `rememberUpdatedState` to ensure the `LaunchedEffect` coroutine uses the latest state for `context`, `view`, `snackbarHostState`, `hostBoundsState`, `onNavigateBack`, without keying them for the `LaunchedEffect`. This will create a reference to those value which can be captured and updated: https://developer.android.com/develop/ui/compose/side-effects#rememberupdatedstate

```kotlin
  val latestContext by rememberUpdatedState(context)
  val latestView by rememberUpdatedState(view)
  val latestSnackbarHostState by rememberUpdatedState(snackbarHostState)
  val latestHostBoundsState by rememberUpdatedState(hostBoundsState)
  val latestOnNavigateBack by rememberUpdatedState(onNavigateBack)
  val latestLaunchRoleRequest by rememberUpdatedState(defaultSmsRoleLauncher::launch)

  LaunchedEffect(screenModel) {
      screenModel.effects.collect { effect ->
          screenModel.handleConversationScreenEffect(
              context = latestContext,
              view = latestView,
              snackbarHostState = latestSnackbarHostState,
              hostBoundsState = latestHostBoundsState,
              effect = effect,
              launchRoleRequest = latestLaunchRoleRequest,
              onNavigateBack = latestOnNavigateBack,
              onDraftSent = { draftSentTick.intValue++ },
          )
      }
  }

```

src/com/android/messaging/ui/conversation/composer/ui/ConversationComposeMessageField.kt
```diff
@@ -0,0 +1,322 @@
...
+    )
+}
+
+@Composable
+private fun ConversationComposePlaceholder() {
+    Text(
+        text = stringResource(id = R.string.compose_message_view_hint_text),
+        style = MaterialTheme.typography.bodyLarge,
+    )
+}
```
inthewaves, line 190: I think we should show the name of the currently selected SIM here like old Messages app

inthewaves (changes requested):

src/com/android/messaging/ui/conversation/messages/model/message/ConversationMessageUiModel.kt
```diff
@@ -0,0 +1,78 @@
...
+import androidx.compose.runtime.Immutable
+import androidx.compose.runtime.Stable
+import com.android.messaging.datamodel.data.ParticipantData
+
+@Immutable
+internal data class ConversationMessageUiModel(
+    val messageId: String,
+    val conversationId: String,
+    val text: String?,
+    val parts: List<ConversationMessagePartUiModel>,
```
inthewaves: Class is marked immutable but uses a plain `List`

src/com/android/messaging/ui/conversation/screen/ConversationSimSheetState.kt
```diff
@@ -0,0 +1,58 @@
...
+import androidx.compose.runtime.saveable.Saver
+import androidx.compose.runtime.saveable.rememberSaveable
+import androidx.compose.runtime.setValue
+
+@Stable
+internal class ConversationSimSheetState {
+    var isVisible: Boolean by mutableStateOf(value = false)
+        private set
+
+    internal var isAvailable: Boolean = false
```
inthewaves: Strictly speaking, `internal var isAvailable: Boolean = false` would make this class [not eligible for `@Stable` normally](https://github.com/JetBrains/kotlin/blob/31660fffadd103332673e3b5c358d6591ce4cb06/plugins/compose/compiler-hosted/src/main/java/androidx/compose/compiler/plugins/kotlin/analysis/Stability.kt#L389) but the annotation forces it to be. I guess this technically isn't an issue right now, because `isAvailable` is not being read by composition

src/com/android/messaging/ui/conversation/composer/ui/ConversationComposeMessageField.kt
RankoR: Old UX/UI makes input taller, which is not good IMO. Copied Google Messages indication:

image

res/layout/attachment_preview.xml
```diff
@@ -1,44 +0,0 @@
...
-    android:scrollbars="none">
-
-    <FrameLayout
-        android:id="@+id/attachment_content"
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        android:layout_gravity="right">
-        <FrameLayout
-            android:id="@+id/attachment_view"
-            style="@style/AttachmentPreviewAttachmentStyle"
```
m4pl, line 30: This style can also be removed from values/styles.xml and values-ldrtl/styles.xml

res/layout/attachment_preview.xml
```diff
@@ -1,44 +0,0 @@
...
-            android:layout_width="match_parent"
-            android:layout_height="wrap_content"
-            android:layout_marginTop="10dp" />
-
-        <ImageButton
-            android:id="@+id/close_button"
-            style="@style/AttachmentPreviewCloseButtonStyle"
```
m4pl, line 37: The same situation is here

res/layout/attachment_preview.xml
```diff
@@ -1,44 +0,0 @@
...
-            android:layout_width="wrap_content"
-            android:layout_height="wrap_content"
-            android:background="@null"
-            android:src="@drawable/ic_remove_light"
```
m4pl, line 41: Also this resource

res/layout/compose_message_view.xml
```diff
@@ -1,187 +0,0 @@
...
-        android:orientation="vertical">
-
-        <View
-            android:id="@+id/top_margin"
-            android:layout_width="match_parent"
-            android:layout_height="16dp" />
-
-        <include layout="@layout/attachment_preview"
-            android:id="@+id/attachment_draft_view"
-            style="@style/ComposeMessageViewDraftViewStyle"
```
m4pl, line 45: Same is here

res/layout/compose_message_view.xml
```diff
@@ -1,187 +0,0 @@
...
-            android:layout_width="match_parent"
-            android:layout_height="0dp"
-            android:layout_marginBottom="4dp"
-            android:layout_weight="1"
-            android:visibility="gone" />
-
-        <RelativeLayout
-            android:id="@+id/subject_view"
-            style="@style/SubjectViewStyle"
-            android:paddingTop="@dimen/compose_message_subject_top_padding"
```
m4pl, line 55: Please make sure to clean up any other remaining styles, colors, and dimensions across the project that are no longer needed after this refactoring

res/menu/conversation_menu.xml
```diff
@@ -1,56 +0,0 @@
...
-    limitations under the License.
--->
-<menu xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools"
-    xmlns:appcompat="http://schemas.android.com/apk/res-auto"
-    tools:context="com.android.messaging.ConversationActivity" >
-
-    <item
-        android:id="@+id/action_call"
-        android:icon="@drawable/ic_phone_small_light"
```
m4pl, line 24: Also this icon

res/values/strings.xml
```diff
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!--
+<?xml version="1.0" encoding="utf-8"?><!--
```
m4pl:
```suggestion
<?xml version="1.0" encoding="utf-8"?>
<!--
```

src/com/android/messaging/data/contact/repository/ContactsRepository.kt
```diff
@@ -0,0 +1,573 @@
...
+        val emailRows = readDestinationRowsForContacts(
+            uri = Email.CONTENT_URI,
+            kind = ContactDestination.Kind.EMAIL,
+            contactIds = contactIds,
+        )
+
+        val rowsByContact = LinkedHashMap<Long, MutableList<DestinationRow>>()
+        contactIds.forEach { contactId ->
+            rowsByContact[contactId] = mutableListOf()
+        }
```
m4pl:
```suggestion
        val rowsByContact = contactIds.associateWith {
            mutableListOf<DestinationRow>()
        }
```

inthewaves (changes requested):
Message animations appears to be a bit janky, I think we'll be good after this:

https://github.com/user-attachments/assets/603c3677-a578-468e-9310-b518db5ddef3

src/com/android/messaging/ui/conversation/addparticipants/AddParticipantsScreen.kt
```diff
@@ -0,0 +1,212 @@
...
+                is AddParticipantsEffect.NavigateToConversation -> {
+                    onNavigateToConversation(effect.conversationId)
+                }
+
+                is AddParticipantsEffect.ShowMessage -> {
+                    UiUtils.showToastAtBottom(effect.messageResId)
+                }
+            }
+        }
+    }
```
inthewaves: Use `rememberUpdatedState` so that we don't key by `onNavigateToConversation`

inthewaves (changes requested):

src/com/android/messaging/data/contact/repository/ContactsRepository.kt
```diff
@@ -0,0 +1,573 @@
...
+
+    private fun groupRowsByContactPreservingOrder(
+        rows: List<DestinationRow>,
+    ): List<List<DestinationRow>> {
+        val rowsByContact = LinkedHashMap<Long, MutableList<DestinationRow>>()
+        rows.forEach { row ->
+            rowsByContact.getOrPut(row.contactId) { mutableListOf() }.add(row)
+        }
+
+        return rowsByContact.values.toList()
```
m4pl:
```suggestion
        val rowsByContact = rows.groupBy(DestinationRow::contactId)
        return rowsByContact.values.toList()
```

src/com/android/messaging/ui/UIIntents.java
m4pl: No callers left after the rewrite

src/com/android/messaging/data/contact/model/Contact.kt
```diff
@@ -0,0 +1,13 @@
+package com.android.messaging.data.contact.model
+
+import androidx.compose.runtime.Immutable
+import kotlinx.collections.immutable.ImmutableList
+
+@Immutable
```
m4pl: Compose's `@Immutable` leaking into a data layer model. I'd keep the data class plain and put `@Immutable` on the UI wrapper

src/com/android/messaging/ui/UIIntents.java
m4pl: Noticed `launchAttachmentChooserActivity` isn't called anywhere after the rewrite. If that's intentional, the chooser stack (`AttachmentChooserActivity / Fragment`, `AttachmentGridView`, `AttachmentPreviewFactory`, `MultiAttachmentLayout`, the matching layouts and the manifest entry) becomes dead code. Same goes for the strings/dimens that only those layouts reference.
Not asking to do it in this PR - it pulls in a lot of unrelated deletions and would bloat the diff. A follow-up PR is probably cleaner. Just flagging so it doesn't get forgotten

src/com/android/messaging/data/conversation/model/metadata/ConversationComposerAvailability.kt
```diff
@@ -0,0 +1,31 @@
+package com.android.messaging.data.conversation.model.metadata
+
+internal data class ConversationComposerAvailability(
```
m4pl: A sealed interface would fit better here - right now there are too many possible combinations of the booleans, but only a couple of them make sense

src/com/android/messaging/data/conversation/mapper/ConversationMessageDataDraftMapper.kt
```diff
@@ -0,0 +1,86 @@
...
+                null
+            }
+        }
+    }
+
+    private fun normalizePartDimension(size: Int): Int? {
+        return size.takeIf { it != MessagePartData.UNSPECIFIED_SIZE }
+    }
+
+    private val String.isPhotoPickerUri: Boolean
```
m4pl: I think a regular function like `private fun isPhotoPickerUri(uri: String)` would be more readable here

src/com/android/messaging/data/conversation/repository/ConversationsRepository.kt
```diff
@@ -0,0 +1,377 @@
...
+    private fun getConversationMessageData(
+        conversationId: String,
+        messageId: String,
+    ): ConversationMessageData? {
+        if (conversationId.isBlank() || messageId.isBlank()) {
+            return null
+        }
+
+        return when {
+            conversationId.isBlank() || messageId.isBlank() -> null
```
m4pl: The blank check is duplicated - there's an early return on top and the same condition inside the `when` below, which is unreachable. I think the `when` can be removed.

src/com/android/messaging/data/media/model/ConversationMediaItem.kt
```diff
@@ -0,0 +1,16 @@
+package com.android.messaging.data.media.model
+
+internal data class ConversationMediaItem(
+    val mediaId: String,
+    val contentUri: String,
+    val contentType: String,
+    val mediaType: ConversationMediaType,
```
m4pl: `mediaType` is set but never read anywhere. I think it should either be used or removed

src/com/android/messaging/ui/conversation/mediapicker/delegate/ConversationMediaPickerDelegate.kt
```diff
@@ -0,0 +1,406 @@
...
+    @param:DefaultDispatcher
+    private val defaultDispatcher: CoroutineDispatcher,
+) : ConversationMediaPickerDelegate {
+
+    private val _effects = MutableSharedFlow<ConversationScreenEffect>(
+        extraBufferCapacity = 1,
+    )
+    private val photoPickerAttachmentLock = Any()
+
+    private val pendingAttachmentJobs = mutableMapOf<String, Job>()
```
m4pl: I couldn't find a place where `pendingAttachmentJobs` is populated - only reads and clears. Looks like dead state, worth a second look

src/com/android/messaging/domain/conversation/usecase/draft/GetConversationDraftSendProtocol.kt
```diff
@@ -0,0 +1,78 @@
...
+        val groupConversationRequiresMms = conversationMetadata.isGroupConversation &&
+            MmsUtils.groupMmsEnabled(selfSubId)
+
+        val emailAddressRequiresMms = MmsSmsUtils.getRequireMmsForEmailAddress(
+            conversationMetadata.includeEmailAddress,
+            selfSubId,
+        )
+
+        return when {
+            draft.attachments.isNotEmpty() -> true
```
m4pl, line 52: I think this would be clearer as a single `||` expression instead of a `when` where most branches return `true`

res/layout/compose_message_view.xml
RankoR: Removed everything that is safe to remove. Will clean up the leftovers after finishing transition to Compose.

src/com/android/messaging/domain/conversation/usecase/draft/GetConversationDraftSendProtocol.kt
RankoR: Too long to be `||` expression in my opinion - `when` is easier to read with such number of conditions.

app/src/androidTest/java/com/android/messaging/ui/conversation/metadata/ui/ConversationTopAppBarTest.kt
```diff
@@ -0,0 +1,52 @@
...
+    return ConversationMetadataUiState.Present(
+        title = "+372 5440 0024",
+        selfParticipantId = "self-participant-id",
+        avatar = ConversationMetadataUiState.Avatar.Single(photoUri = null),
+        participantCount = 1,
+        otherParticipantDisplayDestination = "+372 5440 0024",
+        otherParticipantPhoneNumber = "+37254400024",
+        otherParticipantContactLookupKey = null,
+        isArchived = false,
+        composerAvailability = ConversationComposerAvailability.editable(),
```
inthewaves: Test compilation error


```suggestion
        composerAvailability = ConversationComposerAvailability.Editable,
```

inthewaves (approved):
LGTM, good work, very complex screen. Just need to fix a test compile error (all 12 tests passed for me otherwise)

PR #112, opened 2026-05-19
Conversation Settings Screen: Java/XML to Kotlin/Compose migration

m4pl:
Migrated the Conversation Settings screen from Java + XML to Kotlin + Compose. Partially reused code from [PR](https://github.com/GrapheneOS/Messaging/pull/109) to avoid conflicts and code duplication. Additionally, implemented logic for blocking, archiving, and snooze settings.

**Note:** This PR is partially blocked by [Rewrite Conversation screen PR](https://github.com/GrapheneOS/Messaging/pull/109), as the SIM card switching logic in chat settings cannot be properly tested until it is merged.

### Screenshots

| Before | After |
|--------|-------|
|Screenshot_20260520_001018|Screenshot_20260519_235539|

### Video
[Screen_recording_20260520_012114.webm](https://github.com/user-attachments/assets/4ac0aeeb-56c3-4e48-ab56-7ce7027d8d4c)


src/com/android/messaging/ui/conversationsettings/common/ConversationSettingsTokens.kt
```diff
@@ -0,0 +1,33 @@
...
+internal val GroupedMiddleItemShape = RoundedCornerShape(
+    size = GroupedItemInnerCornerRadius,
+)
+
+internal val GroupedBottomItemShape = RoundedCornerShape(
+    topStart = GroupedItemInnerCornerRadius,
+    topEnd = GroupedItemInnerCornerRadius,
+    bottomStart = SettingsCardCornerRadius,
+    bottomEnd = SettingsCardCornerRadius,
+)
```
inthewaves: I think it would be better to derive these from  MaterialTheme shapes, something like (check the default `dp`s of the shapes)

```
val MaterialTheme.settingsCardShape: CornerBasedShape
    @Composable @ReadOnlyComposable
    get() = shapes.extraLarge

val MaterialTheme.groupedMiddleItemShape: CornerBasedShape
    @Composable @ReadOnlyComposable
    get() = shapes.extraSmall

val MaterialTheme.groupedTopItemShape: CornerBasedShape
    @Composable @ReadOnlyComposable
    get() = shapes.extraLarge.copy(
        bottomStart = shapes.extraSmall.bottomStart,
        bottomEnd = shapes.extraSmall.bottomEnd,
    )

val MaterialTheme.groupedBottomItemShape: CornerBasedShape
    @Composable @ReadOnlyComposable
    get() = shapes.extraLarge.copy(
        topStart = shapes.extraSmall.topStart,
        topEnd = shapes.extraSmall.topEnd,
    )
```

app/src/androidTest/java/com/android/messaging/ui/conversationsettings/screen/ConversationSettingsNavigationTest.kt
```diff
@@ -0,0 +1,79 @@
...
+        verify(exactly = 1) {
+            screenModel.setConversationId(PARTICIPANT_CONVERSATION_ID)
+        }
+    }
+
+    @Test
+    fun closeAfterBlock_fromRoot_navigatesBackWithFinishResult() {
+        renderScreen()
+
+        emitNavEvent(ConversationSettingsNavEvent.CloseAfterBlock)
```
inthewaves: Compile error `Unresolved reference 'CloseAfterBlock'`

src/com/android/messaging/data/conversationsettings/model/SnoozeOption.kt
```diff
@@ -0,0 +1,16 @@
...
+private const val TWENTY_FOUR_HOURS = 24L
+
+internal enum class SnoozeOption(
+    val durationMillis: Long,
+) {
+    OneHour(TimeUnit.HOURS.toMillis(ONE_HOUR)),
+    EightHours(TimeUnit.HOURS.toMillis(EIGHT_HOURS)),
+    TwentyFourHours(TimeUnit.HOURS.toMillis(TWENTY_FOUR_HOURS)),
+    Always(Long.MAX_VALUE),
+}
```
inthewaves, line 13: For Kotlin, I think Kotlin Duration is more idomatic than Java TimeUnit

src/com/android/messaging/ui/conversationsettings/screen/ConversationSettingsViewModel.kt
```diff
@@ -0,0 +1,214 @@
...
+                emitNavigationEvent(NavEvent.CloseAfterArchive)
+            }
+
+            is Action.UnblockClicked -> {
+                delegate.setDestinationBlocked(false)
+            }
+
+            is Action.BlockConfirmed -> {
+                delegate.setDestinationBlocked(true)
+            }
```
inthewaves, line 98: Going by the test compilation failure, for `Action.BlockConfirmed`, are we missing a `CloseAfterBlock` here? It seems old Messaging fragment does close after block:

https://github.com/GrapheneOS/Messaging/blob/a97359e156595630d7d45dfe16d6c28f5deb3c1e/src/com/android/messaging/ui/conversationsettings/PeopleAndOptionsFragment.java#L140-L163

AndroidManifest.xml
```diff
@@ -230,17 +230,24 @@
...
+            android:theme="@style/Theme.Compose">
             <meta-data
                 android:name="android.support.PARENT_ACTIVITY"
                 android:value="com.android.messaging.ui.conversation.ConversationActivity" />
         </activity>
 
+        <activity-alias
+            android:name=".ui.conversationsettings.PeopleAndOptionsActivity"
+            android:exported="true"
+            android:targetActivity=".ui.conversationsettings.ConversationSettingsActivity" />
```
inthewaves: I think we don't need this `activity-alias`. We needed it for the app settings activity because it originally had `android:exported="true"` so it could've potentially broken external apps that used that activity,

For here, the original `PeopleAndOptionsActivity` had no `android:exported` attribute (and no intent-filter), so it was effectively not exported

src/com/android/messaging/ui/conversationsettings/screen/ConversationSettingsEffectHandler.kt
```diff
@@ -0,0 +1,81 @@
...
+                UIIntents.get().launchPhoneCallActivity(
+                    activity,
+                    effect.phoneNumber,
+                    Point(0, 0),
+                )
+            }
+
+            is Effect.ShowOrAddContact -> {
+                ContactUtil.showOrAddContact(
+                    View(activity),
```
inthewaves: `ContactUtil.showOrAddContact` forwards the view to `ContactsContract.QuickContact.showQuickContact(context, view, lookupUri, MODE_LARGE, null)`, which uses the view's window attached on-screen location to position the QuickContact card. But `View(activity)` doesn't have it attached to any window

https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/core/java/android/provider/ContactsContract.java;l=10095

https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/core/java/android/provider/ContactsContract.java;l=9988-9992?q=showQuickContact&ss=android

```java
        /**
         * Constructs the QuickContacts intent with a view's rect.
         * @hide
         */
        public static Intent composeQuickContactsIntent(Context context, View target, Uri lookupUri,
                int mode, String[] excludeMimes) {
            // Find location and bounds of target view, adjusting based on the
            // assumed local density.
            final float appScale = context.getResources().getCompatibilityInfo().applicationScale;
            final int[] pos = new int[2];
            target.getLocationOnScreen(pos);
```

Better would be to use https://developer.android.com/reference/kotlin/androidx/compose/ui/platform/package-summary#LocalView() which has the View that Compose is hosted by

src/com/android/messaging/data/conversationsettings/repository/ConversationSettingsRepository.kt
```diff
@@ -0,0 +1,100 @@
...
+        val participants = queryOtherParticipants(conversationId)
+        val metadata = conversationsRepository.getConversationMetadataSnapshot(
+            conversationId = conversationId,
+        )
+
+        return ConversationSettingsData(
+            conversationId = conversationId,
+            conversationTitle = metadata?.conversationName.orEmpty(),
+            isArchived = metadata?.isArchived ?: false,
+            isSnoozed = notificationRepository.isSnoozed(conversationId),
```
inthewaves, line 93: This seems to compute `isSnoozed` once and snapshots it based on `System.currentTimeMillis()`. Technically, snooze time can run out while the screen is active, and it won't update the UI

src/com/android/messaging/ui/conversationsettings/screen/ConversationSettingsViewModel.kt
```diff
@@ -0,0 +1,214 @@
...
+
+            is Action.ParticipantPressed -> {
+                resolveConversation(action.destination, shouldOpenChat = true)
+            }
```
inthewaves: Old Messages opened the contact card on participant press

src/com/android/messaging/ui/conversationsettings/screen/mapper/ConversationSettingsUiStateMapper.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    private fun canCall(
+        participant: ParticipantUiState?,
+        isVoiceCapable: Boolean,
+    ): Boolean {
+        val phoneNumber = participant?.normalizedDestination
+
+        return isVoiceCapable &&
+            !phoneNumber.isNullOrBlank() &&
+            !PhoneNumberUtils.isEmergencyNumber(phoneNumber)
```
inthewaves: marked as deprecated:

> Please use TelephonyManager.isEmergencyNumber(String) instead.

src/com/android/messaging/ui/conversationsettings/screen/mapper/ConversationSettingsUiStateMapper.kt
```diff
@@ -0,0 +1,101 @@
...
+            canShowContact = !otherParticipant?.normalizedDestination.isNullOrBlank(),
+            isContactSaved = isContactSaved(otherParticipant),
+        )
+    }
+
+    private fun canCall(
+        participant: ParticipantUiState?,
+        isVoiceCapable: Boolean,
+    ): Boolean {
+        val phoneNumber = participant?.normalizedDestination
```
inthewaves: This can be an email

https://github.com/GrapheneOS/Messaging/blob/97fc0e8995bdb2093c49e7ddb2ab071627c5036e/src/com/android/messaging/datamodel/data/ParticipantData.java#L179-L182

Can use something like `PhoneNumberUtils.isWellFormedSmsAddress` to make sure

src/com/android/messaging/ui/conversationsettings/screen/ConversationSettingsScreen.kt
```diff
@@ -0,0 +1,661 @@
...
+        screenModel.setConversationId(targetConversationId)
+    }
+
+    LifecycleEventEffect(Lifecycle.Event.ON_RESUME) {
+        screenModel.refreshState()
+    }
+
+    LaunchedEffect(screenModel, effectHandler) {
+        screenModel.effects.collect(effectHandler::handle)
+    }
```
inthewaves: Technically effectHandler is only created in Activity onCreate so it's only created once (and recreating activity would recompose everything anyway). But we generally shouldn't use what's effectively a lambda as a key to a `LaunchedEffect`; it's better to use https://developer.android.com/develop/ui/compose/side-effects#rememberupdatedstate

inthewaves (changes requested):

src/com/android/messaging/ui/conversationsettings/screen/ConversationSettingsViewModel.kt
m4pl: Ah, yes, exactly. Thanks for pointing that out, I forgot to fix the test in the last commit. I used Google's Messages app as a reference: it doesn't go back after a block, but it does after an archive. I decided to replicate that behavior, so I'll fix the test to match it.

src/com/android/messaging/ui/conversationsettings/screen/ConversationSettingsViewModel.kt
m4pl: I changed this because, in my opinion, it makes more sense to open the chat when tapping on a participant. But I can easily change it back if you want. Also, just a heads-up that there is now a `Contact info` button at the top of the settings:
Screenshot_20260521_101041

src/com/android/messaging/data/conversationsettings/repository/ConversationSettingsRepository.kt
m4pl: I agree. It's unlikely, but it could happen. I'll correct it.

inthewaves:
I've approved the conversation screen rewrite (https://github.com/GrapheneOS/Messaging/pull/109), so we'll be able to properly test this more after that

src/com/android/messaging/ui/conversationsettings/screen/ConversationSettingsViewModel.kt
inthewaves: Ah okay, that makes sense. But from a consistency point of view, a lot of other messaging apps will show some sort of contact info or options in this area instead of opening the chat. e.g. the old Messaging app open contact info, Google Messages has a popup with options, Signal and WhatsApp show a bottom sheet with actions

Also, opening chat screen from here can have other implications. e.g., someone might want to leave a message for a person in a group unread because they might not want to read it at that time but still want the unread status so that they know to come back later. If we make this default open the chat screen, it will mark those messages as read. If they are used to other messaging apps showing contact options in this area, then this will be a surprise.

Generally, for actual groups, I think it's better to make contact info or at least some options as the primary action instead of as the secondary action on the side, especially for people they don't have in their contacts yet.

Google Messages: Contact



Google Messages: Non-contact



Signal

m4pl: Thanks for ideas, I'll refine the approach

m4pl: I have added the logic to open the user preview when clicking on the item. I also added this same preview to the blocked contacts screen. I tried to make it look nice and clear:
Screenshot_20260525_101114

src/com/android/messaging/ui/core/Theme.kt
```diff
@@ -19,6 +30,7 @@ fun AppTheme(
 
     MaterialTheme(
         colorScheme = colorScheme,
+        shapes = AppShapes,
         content = content,
```
inthewaves, line 34: cc @RankoR as both of you have been doing stuff with shapes in PRs: Does this look okay?

src/com/android/messaging/ui/conversationsettings/common/ConversationSettingsComponents.kt
```diff
@@ -0,0 +1,387 @@
...
+
+        ParticipantInfo(
+            displayName = participant.displayName,
+            details = participant.details,
+            modifier = Modifier.weight(1f),
+        )
+
+        if (onAction != null) {
+            ParticipantInfoButton(onClick = onAction)
+        }
```
inthewaves, line 221: I think this info button might be redundant now that clicking the row has that quick actions popup

src/com/android/messaging/ui/conversationsettings/screen/ConversationSettingsViewModel.kt
inthewaves: That looks pretty good! (I don't know why this comment showed up down here but it's for https://github.com/GrapheneOS/Messaging/pull/112#discussion_r3300523496)

src/com/android/messaging/ui/common/components/AnchorRelativePositionProvider.kt
```diff
@@ -0,0 +1,48 @@
...
+    override fun calculatePosition(
+        anchorBounds: IntRect,
+        windowSize: IntSize,
+        layoutDirection: LayoutDirection,
+        popupContentSize: IntSize,
+    ): IntOffset {
+        val popupWidth = popupContentSize.width
+        val popupHeight = popupContentSize.height
+
+        val rightX = anchorBoundsPx.left
```
inthewaves: `AnchorRelativePositionProvider` seems to already receive an `anchorBounds` parameter the `calculatePosition` function, so probably don't need to store `anchorBoundsPx` in the class

src/com/android/messaging/ui/conversationsettings/screen/ConversationSettingsScreen.kt
```diff
@@ -0,0 +1,680 @@
...
+                        onAction(ParticipantAction.ParticipantCallClicked(destination))
+                    }.takeIf { participant.canCall },
+                    onContactClick = {
+                        onAction(ParticipantAction.ParticipantContactInfoClicked(participant))
+                    }.takeIf { hasDestination },
+                    onAction = {
+                        onAction(ParticipantAction.ParticipantActionPressed(destination))
+                    }.takeIf { hasDestination && isGroup },
+                )
+            }
```
inthewaves, line 616: Better for us surround this `ParticipantItem` with `key(some stable id from participant)` especially since this is in a `Column` and group participants could technically change, each participant holds its own state + has side effects, etc.

From https://developer.android.com/develop/ui/compose/lifecycle#add-info-smart-recomposition:

> Calling a composable multiple times will add it to Composition multiple times as well. When calling a composable multiple times from the same call site, Compose doesn’t have any information to uniquely identify each call to that composable, so the execution order is used in addition to the call site in order to keep the instances distinct. This behavior is sometimes all that is needed, but in some cases it can cause unwanted behavior.

inthewaves (changes requested):

src/com/android/messaging/ui/core/Theme.kt
m4pl: I copied this code from @RankoR

src/com/android/messaging/ui/conversationsettings/common/ConversationSettingsComponents.kt
m4pl: They actually lead to different screens - the popup's contact action opens the system contact screen, while the info button navigates to our in-app participant info screen. I'd like to keep both behaviors, but happy to hear your thoughts. Maybe we should surface the in-app one differently?

src/com/android/messaging/ui/common/components/AnchorRelativePositionProvider.kt
m4pl: The `anchorBounds` param Compose passes is the bounds of the Popup's parent layout, not the avatar's. Since the popup should be anchored to the avatar specifically, those bounds are measured explicitly via `onGloballyPositioned`. Let me know if you see a cleaner way!

src/com/android/messaging/ui/conversationsettings/screen/ConversationSettingsScreen.kt
m4pl: Done. Added a stable id: String to ParticipantUiState.

src/com/android/messaging/ui/common/components/AnchorRelativePositionProvider.kt
inthewaves: Maybe we could just put the avatar + quick actions inside of an avatar-sized `Box` inside of `ParticipantAvatarSlot`, then the ParticipantAvatar can just use `Modifer.matchParentSize()`

src/com/android/messaging/ui/conversationsettings/common/ConversationSettingsComponents.kt
inthewaves: Ah I see, that seems fine then. Google Messages does seem to do it the opposite way: their in-app button is in the popup but the side button (a vertical 3-dot icon) is the one that opens system contact screen

inthewaves (approved):
Looks good

inthewaves:
@m4pl Let us know when it's ready to review again (currently GitHub says there's conflicts)

m4pl:
@inthewaves, thanks. I've resolved the conflicts and the PR is ready for review

src/com/android/messaging/datamodel/BugleNotifications.java
```diff
@@ -122,7 +124,7 @@ public static void update(final String conversationId, final int coverage) {
         }
     Assert.isNotMainThread();
 
-        if (!shouldNotify()) {
+        if (!shouldNotify(conversationId)) {
             cancel(PendingIntentConstants.SMS_NOTIFICATION_ID);
             return;
         }
```
inthewaves, line 148: we probably should cancel with the conversationId here too instead of cancelling all SMS notifications. When receiving a message for a snoozed conversation, this will cancel all SMS notifications even for other conversations, since no conversation ID is passed

inthewaves (changes requested):

src/com/android/messaging/datamodel/BugleNotifications.java
m4pl: Ah, right. Thanks, I missed that

inthewaves (approved):
LGTM

RankoR (approved):

Issue #120, opened 2026-05-22
Add blocked user indication to the conversation screen


RankoR:
<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/a4b47b9f-264d-4ef2-bbb9-048f7fc18fc4" />

PR #121, opened 2026-05-22
Blocked Participants Screen: Java/XML to Kotlin/Compose migration

m4pl:
Migrated the Blocked Participants screen from Java + XML to Kotlin + Compose. Partially reused code from the Conversation Settings migration [PR](https://github.com/GrapheneOS/Messaging/pull/109) to avoid conflicts and code duplication.

Reworked the data layer: the list is now built from a single SQL query that joins blocked participants with existing 1-on-1 conversations. Only blocked participants who currently have a direct chat are shown - after deleting a blocked chat the row disappears, but the participant stays blocked.

Added multi-select mode in the top bar (counter + clear selection + batched delete) and covered the screen with Compose UI tests.

**Out of scope ([task](https://github.com/GrapheneOS/Messaging/issues/120)):** the chat screen will surface a "this chat is blocked" indicator so the delete/block flow becomes more transparent to the user.

### Screenshots

| Before | After |
|--------|-------|
Screenshot_20260522_170604|Screenshot_20260522_170534

### Video
[Screen_recording_20260522_171007.webm](https://github.com/user-attachments/assets/a4a63522-70fb-43d8-9167-243e727fe7d7)


src/com/android/messaging/ui/common/components/ParticipantQuickActionsPopup.kt
```diff
@@ -0,0 +1,323 @@
...
+    Box(
+        modifier = Modifier
+            .fillMaxWidth()
+            .aspectRatio(1f)
+            .clip(
+                RoundedCornerShape(
+                    topStart = CardCornerRadius,
+                    topEnd = CardCornerRadius,
+                ),
+            ),
```
inthewaves: Let's use shapes based on the the MaterialTheme shapes for this if possible

src/com/android/messaging/ui/common/components/AnchorRelativePositionProvider.kt
```diff
@@ -0,0 +1,48 @@
...
+    private val gapPx: Int,
+    private val transformOriginState: MutableState<TransformOrigin>,
+) : PopupPositionProvider {
+
+    override fun calculatePosition(
+        anchorBounds: IntRect,
+        windowSize: IntSize,
+        layoutDirection: LayoutDirection,
+        popupContentSize: IntSize,
+    ): IntOffset {
```
inthewaves, line 22: We should use a Box as a parent like the conversation settings PR for cleaner code

inthewaves (changes requested):

src/com/android/messaging/ui/blockedparticipants/common/BlockedParticipantItem.kt
```diff
@@ -0,0 +1,272 @@
...
+        )
+
+        Text(
+            text = details.orEmpty(),
+            style = MaterialTheme.typography.bodySmall,
+            color = MaterialTheme.colorScheme.onSurfaceVariant,
+            maxLines = 1,
+            overflow = TextOverflow.Ellipsis,
+        )
+    }
```
inthewaves: I think we just have a phone number and no display name, we should vertically center the text

inthewaves (changes requested):

src/com/android/messaging/ui/blockedparticipants/screen/delegate/BlockedParticipantsDelegate.kt
```diff
@@ -0,0 +1,116 @@
...
+        participants: List<BlockedParticipantUiState>,
+    ): ImmutableSet<String> {
+        if (isEmpty()) return this
+
+        val knownIds = participants.mapTo(HashSet(participants.size)) {
+            it.participantId
+        }
+
+        return filter { it in knownIds }.toPersistentSet()
+    }
```
inthewaves: `filter { it in knownIds }` creates an intermediate List

`PersistentSet` has an APIs for this part like `retainAll` (renamed to `retainingAll` in upcoming kotlinx immutable collections 0.5.0): https://kotlin.github.io/kotlinx.collections.immutable/kotlinx-collections-immutable/kotlinx.collections.immutable/-persistent-set/retaining-all.html. This API doesn't exist on `ImmutableSet`

It might be worth changing `selectedParticipantIds` to use `PersistentSet` instead of `ImmutableSet` so that we can use `retainAll` directly.

src/com/android/messaging/ui/blockedparticipants/screen/delegate/BlockedParticipantsDelegate.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    override fun toggleSelection(participantId: String) {
+        _state.update { current ->
+            val updated = if (participantId in current.selectedParticipantIds) {
+                current.selectedParticipantIds - participantId
+            } else {
+                current.selectedParticipantIds + participantId
+            }
+            current.copy(selectedParticipantIds = updated.toPersistentSet())
```
inthewaves: `ImmutableSet` has no `ImmutableSet.minus` or `ImmutableSet.plus` operators, so this will use Kotlin's standard `Set.minus` and `Set.plus`, hence why `val updated` can only be typed as a `Set`. `Set.minus` and `Set.plus` are generic so they will do $O(n)$ copies every time

For example, from the Kotlin standard library:

```kotlin
public operator fun <T> Set<T>.minus(element: T): Set<T> {
      val result = LinkedHashSet<T>(mapCapacity(size))
      var removed = false
      return this.filterTo(result) {
          if (!removed && it == element) {
              removed = true
              false
          } else {
              true
          }
      }
  }

```

It seems it would be better overall to use `PersistentSet` as the type instead of `ImmutableSet` in the state data class. If `selectedParticipantIds` was typed to be a `PersistentSet`, then these operations would use `PersistentSet` (copy-on-write) operations for adding and removing elements which looks more efficient:

`PersistentCollection.plus` and `PersistentCollection.minus` do exist: https://github.com/Kotlin/kotlinx.collections.immutable/blob/master/core/commonMain/src/extensions.kt#L71

`removing` calls `remove` (`removing` seems to be a new API renaming): https://github.com/Kotlin/kotlinx.collections.immutable/blob/master/core/commonMain/src/ImmutableCollection.kt#L91 

`PersistentOrderedSet`'s `remove`, for example, looks efficient (or at least more specialized than creating a brand-new `LinkedHashSet`) and reuses `PersistentHashMap` structure: https://github.com/Kotlin/kotlinx.collections.immutable/blob/master/core/commonMain/src/implementations/persistentOrderedSet/PersistentOrderedSet.kt#L72

inthewaves (changes requested):

src/com/android/messaging/ui/blockedparticipants/common/BlockedParticipantItem.kt
```diff
@@ -0,0 +1,287 @@
...
+                participant = participant,
+                fallbackIcon = fallbackIcon,
+                isSelected = isSelected,
+                onAvatarClick = onAvatarClick,
+                quickActionsVisible = quickActionsVisible,
+                onDismissQuickActions = onDismissQuickActions,
+                onMessageClick = onMessageClick,
+                onCallClick = onCallClick,
+                onContactClick = onContactClick,
+            )
```
inthewaves: When in multi-select mode after a long press, for an unselected contact, tapping their avatar will bring up quick actions instead of selecting that contact/number. For a selected contact, tapping their avatar will do nothing. I think quick actions should be disabled when in multi-select mode

inthewaves (changes requested):

inthewaves (approved):
Looks good but I think this would have to be rebased when conversation settings rewrite https://github.com/GrapheneOS/Messaging/pull/112 is merged

RankoR (approved):

src/com/android/messaging/domain/blockedparticipants/usecase/DeleteDirectChats.kt
```diff
@@ -0,0 +1,30 @@
...
+            .toList()
+
+        if (destinations.isEmpty()) return
+
+        val conversationIds = repository.findDirectConversationIds(destinations)
+
+        conversationIds.forEach { conversationId ->
+            DeleteConversationAction.deleteConversation(conversationId, Long.MAX_VALUE)
+        }
+    }
```
inthewaves: This re-lookup seems unnecessary since the `BlockedParticipantUiState` state already has `conversationId`s

src/com/android/messaging/ui/blockedparticipants/screen/model/BlockedParticipantsUiState.kt
```diff
@@ -0,0 +1,28 @@
...
+internal data class BlockedParticipantsUiState(
+    val isLoading: Boolean = true,
+    val participants: ImmutableList<BlockedParticipantUiState> = persistentListOf(),
+    val selectedParticipantIds: PersistentSet<String> = persistentSetOf(),
+)
+
+@Immutable
+internal data class BlockedParticipantUiState(
+    val participantId: String,
+    val conversationId: String,
```
inthewaves, line 19: UI state already has `participantId` + `conversationId` here

inthewaves (changes requested):

inthewaves (approved):

PR #122, opened 2026-05-26
License Screen: Java/XML to Kotlin/Compose migration

m4pl:
Migrated the Open Source Licenses screen from Java + XML to Kotlin + Compose. Added a script for generating the licenses HTML - see [PR](https://github.com/GrapheneOS/app-docs/pull/3). Also updated libraries from the [Conversation screen PR](https://github.com/GrapheneOS/Messaging/pull/109) to avoid conflicts and code duplication.

### Screenshots

| Before | After |
|--------|-------|
|Screenshot_20260526_143747|Screenshot_20260526_143839|


app/build.gradle.kts
```diff
@@ -121,6 +132,13 @@ android {
 
 dependencies {
     implementation(libs.androidx.appcompat)
+    implementation(libs.androidx.camera.camera2)
+    implementation(libs.androidx.camera.compose)
+    implementation(libs.androidx.camera.core)
+    implementation(libs.androidx.camera.lifecycle)
+    implementation(libs.androidx.camera.video)
+    implementation(libs.androidx.paging.compose)
+    implementation(libs.androidx.paging.runtime)
```
inthewaves, line 142: Perhaps we need to incorporate the license generation script as a Gradle plugin so that it's auto-generated for us, e.g. these dependencies are not in the .html, and some of the versions don't match the actual versions in libs.versions.toml, e.g. org.jetbrains.kotlinx:kotlinx-coroutines-android:1.10.2 is in the html, but the app actual uses version `coroutines = "1.11.0"`

inthewaves (changes requested):

app/build.gradle.kts
m4pl: Thanks for catching this. I forgot to refresh the dependency cache after bumping versions in `libs.versions.toml`, so the script regenerated the HTML against stale input.
Agreed on moving this into Gradle. I'll try to replace the script with a Gradle task in this PR.

m4pl: Done - added a Gradle plugin under `buildSrc/` (`id("messaging.licenses")`). Generation is now `./gradlew :app:generateLicenses`.

Also regenerated `assets/licenses.html` against the current resolution graph.

Note: I leaned on AI for the implementation, but went through every file myself and verified the output after each step.

app/build.gradle.kts
```diff
@@ -3,6 +3,7 @@ import java.io.FileInputStream
 import java.util.Properties
 
 plugins {
+    id("messaging.licenses")
```
inthewaves, line 6: missing licenses for non-gradle-downloaded/pom dependencies such as GIFLIB (lib/platform_external_giflib) and the AOSP modules like platform_frameworks_opt_chips

These were in the previous license.html

https://github.com/GrapheneOS/Messaging/blob/3555b6a4993fc6be3abc0192ce328e8b84182eab/assets/licenses.html#L644

https://github.com/GrapheneOS/Messaging/blob/3555b6a4993fc6be3abc0192ce328e8b84182eab/assets/licenses.html#L12

buildSrc/src/main/kotlin/messaging/licenses/SpdxPolicy.kt
```diff
@@ -0,0 +1,122 @@
...
+        "AGPL-",
+    )
+
+    private val SOURCE_AVAILABILITY = setOf(
+        "LGPL-2.1-only",
+        "LGPL-2.1-or-later",
+        "LGPL-3.0-only",
+        "LGPL-3.0-or-later",
+        "MPL-1.1",
+        "MPL-2.0",
```
inthewaves, line 23: The license for OkHttp 4.12.0 includes this notice at the bottom, although the pom for it is just Apache 2.0. Maybe we need to handle this as part of the source available list

```
-----
NOTICE:

Note that publicsuffixes.gz is compiled from The Public Suffix List:
https://publicsuffix.org/list/public_suffix_list.dat

It is subject to the terms of the Mozilla Public License, v. 2.0:
https://mozilla.org/MPL/2.0/
```

app/build.gradle.kts
inthewaves, line 6: I think we probably also would need a process to make sure license html is up-to-date, maybe a GitHub workflow can regenerate (from a fresh gradle build) and fail if different from checked-in license html or something

inthewaves (changes requested):

buildSrc/src/main/kotlin/messaging/licenses/SpdxPolicy.kt
m4pl: No special handling needed. OkHttp itself is Apache-2.0; the MPL-2.0 only covers the bundled `publicsuffixes.gz`, which we don't modify. For an unmodified file the only MPL obligation is preserving its notice, which the NOTICE block already does, and it links to the upstream source.

buildSrc/src/main/kotlin/messaging/licenses/SpdxPolicy.kt
```diff
@@ -0,0 +1,122 @@
+package messaging.licenses
+
+import org.gradle.api.GradleException
+
+internal object SpdxPolicy {
```
RankoR, line 5: EUPL is missing. It's not popular, but worth adding just in case.

inthewaves (approved):
LGTM

PR #124, opened 2026-05-28
Permission Check Screen: Java/XML to Kotlin/Compose migration

m4pl:
Migrated the Permission Check screen from Java + XML to Kotlin + Compose.

### Screenshots

| Before | After |
|--------|-------|
|Screenshot_20260528_194035|Screenshot_20260528_194003|


res/values/strings.xml
inthewaves: I think these instructions are outdated now (triggered by denying to set it as default SMS app, then denying again with Don't ask again checked). I tried to only grant SMS , Phone, and Contacts without the SMS role but the app still doesn't let me in. It requires Settings > Apps > Messaging > SMS app to set Messaging as the default SMS app

Think we need new strings to say that it needs to be set as the default SMS app. Also, as a nit, it might be better to use a string template here for the app name instead of hardcoding Messaging (more helpful for dev stuff since it's supposed to be "Messaging d" for debug/perf builds)

These existing strings still might be useful, because even when set as the default SMS app / in the SMS role, you can still deny the required permissions. So we probably need to distinguish between two states: one where SMS role is not set to this app, and one where the permissions are missing. The states aren't mutually exclusive, but the SMS role state should be prioritized

inthewaves (changes requested):

inthewaves (approved):

PR #125, opened 2026-06-02
Add tests for conversation screen

RankoR:
Last (of 3) PR with Conversation screen rewritten with Kotlin and Compose.

Adds:

- Unit and Instrumented tests. Used Robolectric unit tests where possible
- Coverage verification and enforcement

Coverage:

- Unit: 93% instructions, 74% branches
- Instrumented: 88% instructions, 62% branches

Doesn't change behavior of production code. All production code changes are:

- Added tags for tests
- Removed one unused method


app/src/test/kotlin/com/android/messaging/domain/conversation/usecase/draft/SendConversationDraftImplTest.kt
```diff
@@ -0,0 +1,613 @@
...
+            )
+        } returns sendData
+        return repository
+    }
+
+    private fun createSubscriptionsRepositoryMock(): SubscriptionsRepository {
+        val repository = mockk<SubscriptionsRepository>(relaxed = true)
+        every { repository.resolveAttachmentLimit() } returns Int.MAX_VALUE
+        return repository
+    }
```
inthewaves: Think we should also test reachable attachment limits as well

app/src/test/kotlin/com/android/messaging/ui/conversation/composer/ui/ConversationComposeBarTest.kt
```diff
@@ -0,0 +1,828 @@
...
+            assertEquals(ConversationAudioRecordingPhase.Idle, audioRecording.phase)
+        }
+    }
+
+    private fun setContent(
+        audioRecording: ConversationAudioRecordingUiState = ConversationAudioRecordingUiState(),
+        messageText: String,
+        subjectText: String = "",
+        sendProtocol: ConversationDraftSendProtocol = ConversationDraftSendProtocol.SMS,
+        segmentCounter: ConversationSegmentCounterUiState? = null,
```
inthewaves, line 764: We should probably have some tests for the segment counter

inthewaves (changes requested):

.github/workflows/build.yml
```diff
@@ -1,20 +1,185 @@
...
+          api-level: 36
+          arch: x86_64
+          target: default
+          disable-animations: true
+          script: >
+            ./gradlew :app:connectedDebugAndroidTest :app:jacocoAndroidTestVerification
+            -PandroidTestCoverage=true
+            -PandroidTestMinCoverage=80
+            -PandroidTestMinBranchCoverage=50
+            --no-daemon --stacktrace --console=plain
```
m4pl, line 196: What do you think about caching the AVD snapshot? Could shave off the cold emulator boot each run

app/src/test/kotlin/com/android/messaging/ui/conversation/screen/route/ConversationScreenRouteEffectsTest.kt
```diff
@@ -0,0 +1,343 @@
...
+        )
+        composeTestRule.waitForIdle()
+
+        composeTestRule.runOnIdle {
+            verify(exactly = 1) {
+                screenModel.model.onSimSelected(selfParticipantId = SELF_PARTICIPANT_ID)
+            }
+            assertEquals(1, selfParticipantConsumedCount)
+        }
+    }
```
inthewaves: note that in practice this doesn't select the SIM

app/src/test/kotlin/com/android/messaging/ui/conversation/composer/mapper/ConversationComposerUiStateMapperImplTest.kt
```diff
@@ -0,0 +1,473 @@
...
+            areSubscriptionsLoaded = true,
+        )
+
+        assertEquals(matchingSubscription, uiState.simSelector.selectedSubscription)
+        assertEquals(subscriptions, uiState.simSelector.subscriptions)
+        assertTrue(uiState.simSelector.isAvailable)
+    }
+
+    @Test
+    fun map_fallsBackToFirstSubscriptionWhenDraftSelfParticipantIdDoesNotMatch() {
```
inthewaves: We probably shouldn't be having this behavior of falling back to the first SIM. If the default SMS SIM in the OS Settings is set up to be the second SIM, then the actual send will use the second SIM but the UI will display the first SIM.

See https://github.com/GrapheneOS/Messaging/issues/142

inthewaves (changes requested):

app/jacoco-rules/common/excluded-class-patterns.txt
```diff
@@ -0,0 +1,19 @@
...
+**/_MembersInjector*
+
+# Compose and data-holder noise.
+**/*ComposableSingletons*
+**/model/**
+
+# Kotlin compiler-generated classes.
+**/*$DefaultImpls*
+**/*$invokeSuspend$*
+**/*$serializer
```
m4pl: This pattern looks like it may not match the actual Kotlin serialization classes. The compiled files are named like `ConversationNavKey$$serializer.class`, while this rule is `**/*$serializer` without the second `$` or `.class` suffix. Should this be changed to something like `**/*$$serializer.class`?

src/com/android/messaging/ui/conversation/composer/ui/ConversationComposeBar.kt
```diff
@@ -641,7 +640,7 @@ private fun SegmentCounterIndicator(
     Text(
         modifier = Modifier
             .padding(bottom = 4.dp)
-            .clearAndSetSemantics {
+            .semantics {
```
m4pl: This change may affect accessibility behavior. `clearAndSetSemantics` was hiding the raw counter text and exposing only the custom content description, while `semantics` may expose both. Was this change needed only for the test assertion?

app/src/androidTest/kotlin/com/android/common/test/helpers/ShellCommandHelper.kt
```diff
@@ -0,0 +1,152 @@
...
+    private fun shellSingleQuoted(value: String): String {
+        return "'${value.replace(oldValue = "'", newValue = "'\\''")}'"
+    }
+
+    private fun shellWord(value: String): String {
+        return shellSingleQuoted(value = value)
+    }
+
+    private const val SMS_ROLE_NAME = "android.app.role.SMS"
+    private const val SMS_ROLE_RESTORE_DELAY_SECONDS = 15
```
m4pl, line 193: The SMS role restore runs asynchronously after 15 seconds. Could this affect the next instrumentation test if it starts before the role is restored?

app/src/test/kotlin/com/android/messaging/ui/conversation/attachment/mapper/ConversationVCardAttachmentUiModelMapperImplTest.kt
m4pl, line 1: Nit: this test lives under `messages/mapper`, but the mapper under test is in `ui/conversation/attachment/mapper`. Moving it to the matching package/path would make ownership clearer

app/src/test/kotlin/com/android/messaging/ui/conversation/metadata/delegate/ConversationMetadataDelegateImplTest.kt
```diff
@@ -0,0 +1,295 @@
...
+                        )
+                }
+            } finally {
+                harness.cancel()
+            }
+        }
+    }
+
+    @Test
+    fun commandMethods_doNothingWhenConversationIdIsBlankOrNull() {
```
m4pl, line 200: This test says the commands should do nothing for blank/null conversation ids, but it only checks effects and confirmation state

app/src/test/kotlin/com/android/messaging/ui/conversation/composer/mapper/ConversationComposerUiStateMapperImplTest.kt
RankoR: Will fix in a separate PR.

inthewaves:
This branch has conflicts that must be resolved

inthewaves (approved):
Could switch to Android 17 (SDK 37) emulator target in the future

PR #127, opened 2026-06-03
Fix out-of-bounds read in MMS PDU content type parsing

thomasbuilds:


thestinger:
Fixed by https://github.com/GrapheneOS/Messaging/pull/130. Thanks for the report.

src/com/android/messaging/sms/MmsUtils.java
```diff
@@ -2038,7 +2038,12 @@ public static DatabaseMessages.MmsMessage processReceivedPdu(final Context conte
         // Get raw PDU push-data from the message and parse it
         final PduParser parser = new PduParser(pushData,
                 MmsConfig.get(subId).getSupportMmsContentDisposition());
-        final GenericPdu pdu = parser.parse();
+        GenericPdu pdu = null;
+        try {
+            pdu = parser.parse();
+        } catch (final RuntimeException e) {
+            LogUtil.e(TAG, "processReceivedPdu: Failed to parse PUSH data", e);
+        }
```
thomasbuilds, line 2046: @RankoR I've seen this was left out.

`parseValueLength()` throws `RuntimeException("Value length > LENGTH_QUOTE!")` for any value-length octet >= 32, and is called unguarded in `parseHeaders()` for FROM, EXPIRY (a mandatory m-notification-ind header), and others. It throws out of `parse()` before mandatory-field validation, and `processReceivedPdu()` doesn't catch it, so that path still crashes.

Repro (4 bytes): `8C 82 89 40` (m-notification-ind, FROM `0x89`, value-length `0x40`) throws from `parse()` on current main.

`parsePduForAnyCarrier()` already wraps `parse()` in `catch (RuntimeException)` - mirroring it in `processReceivedPdu()` closes the whole class.

PR #128, opened 2026-06-03
Bound GIF dimensions before native transcoding

thomasbuilds:
The native GIF transcoder (`jni/GifTranscoder.cpp`) sizes its render buffers from the GIF's logical screen dimensions, computed as `int * int`:

```cpp
std::vector<GifByteType> srcBuffer(gifIn->SWidth * gifIn->SHeight);
std::unique_ptr<ColorARGB[]> renderBuffer(new ColorARGB[gifIn->SWidth * gifIn->SHeight]);
```

`SWidth`/`SHeight` come straight from the GIF (each up to 65535), and `canBeTranscoded()` only enforced a minimum size (>= 100x100), no maximum. A crafted GIF with large dimensions overflows the multiplication (65535*65535 wraps negative, becoming a huge `size_t`) or just requests an enormous allocation, so `std::vector`/`new[]` throws an uncaught C++ exception (`std::length_error`/`bad_alloc`) that propagates out of JNI and aborts the process (`SIGABRT`). Transcoding is enabled by default and runs on the send/forward path, so sending or forwarding such a GIF crashes Messaging.

### DIFF
--- a/jni/GifTranscoder.cpp
+++ b/jni/GifTranscoder.cpp
@@ -39,6 +39,8 @@ static const ColorARGB TRANSPARENT = 0x0;
 
 #define MAX_COLOR_DISTANCE (255 * 255 * 255)
 
+static const int64_t kGifMaxArea = 4LL * 1024 * 1024;
+
 #define TAG "GifTranscoder.cpp"
 #define LOGD_ENABLED 0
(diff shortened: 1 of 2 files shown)


RankoR (approved):
Please add tests in future similar PRs.

PR #131, opened 2026-06-03
Catch malformed-header exceptions in PduParser.parse

thomasbuilds:
https://github.com/GrapheneOS/Messaging/pull/127#discussion_r3351686343


RankoR:
Hey @thomasbuilds, thanks for your contribution, but I'm already finalizing another solution for that. That was intentionally skipped because it catches too broad exception.

PR #134, opened 2026-06-04
Bound EXIF parser allocations to the APP1 segment size

thomasbuilds:
The app's EXIF parser (com.android.messaging.util.exif) sizes several heap allocations from attacker-controlled EXIF fields bounded only by Integer.MAX_VALUE: tag component counts (readFullTagValue), the IFD0 offset (mDataAboveIfd0), and the compressed-thumbnail and uncompressed-strip lengths (ExifReader). A crafted JPEG under ~80 bytes can declare a value of up to ~2GB, so new byte[...] / new long[...] throws OutOfMemoryError.

The parser is reached from ImageUtils.getOrientation (new ExifInterface().readExif(...)) when an image attachment is decoded for display. getOrientation catches only IOException and MediaResourceManager catches only Exception, so the OutOfMemoryError escapes and crashes the media-loading worker.

This bounds every such size by mApp1End, the length of the Exif APP1 segment (at most ~64KB), which valid EXIF data never exceeds. Oversized declarations are rejected with IOException before allocating, so the loader needs no catch (OutOfMemoryError).

### DIFF
--- a/src/com/android/messaging/util/exif/ExifParser.java
+++ b/src/com/android/messaging/util/exif/ExifParser.java
@@ -217,7 +217,7 @@ public class ExifParser {
 
         parseTiffHeader();
         long offset = mTiffStream.readUnsignedInt();
-        if (offset > Integer.MAX_VALUE) {
+        if (offset > mApp1End) {
             throw new ExifInvalidFormatException("Invalid offset " + offset);
         }
         mIfd0Position = (int) offset;
@@ -495,6 +495,10 @@ public class ExifParser {
         return (int) mJpegSizeTag.getValueAt(0);
     }
 
+    protected int getApp1End() {
+        return mApp1End;
+    }
+
     private void skipTo(int offset) throws IOException {
         mTiffStream.skipTo(offset);
         while (!mCorrespondingEvent.isEmpty() && mCorrespondingEvent.firstKey() < offset) {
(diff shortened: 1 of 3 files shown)


src/com/android/messaging/util/exif/ExifParser.java
```diff
@@ -561,6 +565,9 @@ private ExifTag readTag() throws IOException, ExifInvalidFormatException {
             if (mDataAboveIfd0 != null
                     && offset < mIfd0Position
                     && dataFormat == ExifTag.TYPE_UNDEFINED) {
+                if (numOfComp < 0 || numOfComp > mApp1End) {
+                    throw new ExifInvalidFormatException("Invalid tag size " + numOfComp);
+                }
                 byte[] buf = new byte[(int) numOfComp];
                 System.arraycopy(mDataAboveIfd0, (int) offset - DEFAULT_IFD0_OFFSET,
```
RankoR, line 575: Would be good to validate offset as well:

```
  if (numOfComp < 0 || numOfComp > mApp1End) {
      throw new ExifInvalidFormatException("Invalid tag size " + numOfComp);
  }
  if (offset < DEFAULT_IFD0_OFFSET || offset + numOfComp > mIfd0Position) {
      throw new ExifInvalidFormatException("Invalid tag offset " + offset);
  }
  byte[] buf = new byte[(int) numOfComp];
  System.arraycopy(mDataAboveIfd0, (int) offset - DEFAULT_IFD0_OFFSET,
          buf, 0, (int) numOfComp);
```

RankoR (changes requested):

RankoR (approved):

PR #138, opened 2026-06-05
Make YouTube link previews opt-in

thomasbuilds:
The conversation screen auto-fetches YouTube thumbnails from img.youtube.com as soon as a conversation containing a YouTube link is viewed, disclosing the viewer's IP address and read time to Google with no opt-out. Gate the feature behind a new YouTube link previews toggle in a Privacy settings sub-screen, disabled by default. The preference is re-read when the conversation screen is foregrounded, so toggling it applies to open conversations without reopening them.


RankoR (changes requested):
@thomasbuilds it should be configurable in settings (off by default), not fully removed

I suggest adding a sub-screen - "Privacy" and add it there as a toggle, with a short description that enabling it may reduce privacy

thomasbuilds:
I don't see any reason to special-case this third-party website?

RankoR:
@thomasbuilds it might be extended to other services in the future, especially if we manage to make RCS work one day. Plus, it's a pre-existing and compact (in terms of LoC) feature that improves UX (for those who want to keep it turned on), there's no reason to remove it if it will be opt-in.

But I agree that having it on by default (especially without the ability to turn it off) is a privacy issue.

RankoR (changes requested):
Production combines the repository preference with conversation messages and forwards it into every mapped message in ConversationMessagesDelegate.kt lines 134–164 (https://github.com/GrapheneOS/Messaging/blob/c359552bbab06ecde93bb829220ffc8386fa63c2/src/com/android/messaging/ui/conversation/messages/delegate/ConversationMessagesDelegate.kt#L134-L164).

But ConversationMessagesDelegateImplTest (https://github.com/GrapheneOS/Messaging/blob/c359552bbab06ecde93bb829220ffc8386fa63c2/app/src/test/kotlin/com/android/messaging/ui/conversation/messages/delegate/ConversationMessagesDelegateImplTest.kt) only adds the new repository mock to the constructor; it does not test this behavior.

Please add a focused coroutine test that binds the delegate, returns false and then true from isYouTubeLinkPreviewsEnabled(), triggers refresh(), and verifies the mapper receives the corresponding value.

The rest looks good.

RankoR (approved):

thestinger:
This can't be merged without a rebase.

RankoR (approved):

src/com/android/messaging/ui/conversation/messages/model/message/ConversationMessageUiModel.kt
```diff
@@ -31,6 +31,7 @@ internal data class ConversationMessageUiModel(
     val canForwardMessage: Boolean,
     val canResendMessage: Boolean,
     val canSaveAttachments: Boolean,
+    val isYouTubePreviewEnabled: Boolean,
```
inthewaves: I don't agree with putting `isYouTubePreviewEnabled` inside of `ConversationMessageUiModel`. `isYouTubePreviewEnabled` is application/screen policy, not message data

- This duplicates the same global value across every message
- This can theoretically make impossible states representable, such as one message having it enabled and another disabled
- Forces unrelated mapper tests and fixtures to provide the preference
- This can trigger a rebuild of the entire message list. The preference is combined before `flatMapLatest`; a value change therefore cancels and recreates `observeConversationMessagesUiState(messages)`. e.g. for conversations containing vCards, that recreates every metadata flow and re-emits its `Loading` state

This would belong better in screen-level state such as `ConversationMessagesUiState.Present(messages, youTubeLinkPreviewsEnabled)` and pass it as an explicit rendering-policy parameter to `ConversationMessages`/`ConversationMessage`/`buildConversationMessageContent`

src/com/android/messaging/ui/appsettings/screen/SettingsViewModel.kt
```diff
@@ -106,38 +156,7 @@ internal class SettingsViewModel @Inject constructor(
...
-
-            is Action.NotificationsClicked -> {
-                emitEffect(Effect.OpenNotificationSettings)
-            }
-
-            is Action.LicensesClicked -> {
-                emitEffect(Effect.OpenLicenses)
-            }
+            else -> Unit
         }
```
inthewaves, line 153: I don't think this refactor into `onSubscriptionAction` is worth it, because when we add a new subscription action, the `else -> Unit` no longer forces the programmer to handle it through a compiler error. `onSubscriptionAction` is still on the sealed type `Action`

The original `onAction` itself remains exhaustive and might encourage programmers to handle it there instead

inthewaves (changes requested):

src/com/android/messaging/ui/appsettings/screen/SettingsViewModel.kt
thomasbuilds: I kept the split because `onAction` is already at detekt's limits on main so adding the new branch inline fails the build on both rules. The subscription actions are now a `sealed interface Subscription` and the inner `when` has no `else` so the compiler error you mentioned is preserved.

inthewaves: Ah okay, that looks fine then, thanks!

inthewaves (approved):

PR #140, opened 2026-06-06
Bound conversation notification people list (fixes #102)

### LINKED ISSUE #102: Messaging App Crashes
When trying to receive a text message (PayPal 2fa), the message never arrives. Opening any type of message causes the app to crash. Deleting a message causes the app to crash. 

Device: Pixel 9(caiman)
OS build: BP4A. 260205.002
APP VERSION: 20000012

[System log _ C 1b37da223d36.txt](https://github.com/user-attachments/files/26480490/System.log._.C.1b37da223d36.txt)

RankoR:
A long-lived conversation eventually crashes the app - receiving, opening, or deleting a message throws `TransactionTooLargeException`: https://github.com/GrapheneOS/Messaging/issues/102

`processAndSend` calls `addPerson` once per message line, so the notification's people list grows without bound until the notification parcel exceeds the ~1 MB Binder limit, after which every `NotificationManager.notify` throws.

Fix is to attach each participant only once: `limitPeople` de-duplicates by person identity and caps the list at `MAX_NOTIFICATION_PEOPLE` (25) as a safety backstop.

Behaviour is otherwise unchanged.

### DIFF
--- a/app/src/androidTest/java/com/android/messaging/datamodel/BugleNotificationsTransactionSizeTest.kt
+++ b/app/src/androidTest/java/com/android/messaging/datamodel/BugleNotificationsTransactionSizeTest.kt
@@ -1,0 +1,114 @@
+package com.android.messaging.datamodel
+
+import android.Manifest
+import android.app.NotificationChannel
+import android.app.NotificationManager
+import android.content.Context
+import android.os.TransactionTooLargeException
+import androidx.core.app.NotificationCompat
+import androidx.core.app.NotificationCompat.MessagingStyle
+import androidx.core.app.NotificationManagerCompat
+import androidx.core.app.Person
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import com.android.messaging.R
+import com.android.messaging.util.NotificationChannelUtil
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+
+/**
+ * Instrumented regression test for issue #102
+ *
+ * Builds a conversation notification the way [BugleNotifications.processAndSend] does: attaching
+ * one bounded [Person] list for all message lines and actually posting it through the real
+ * framework. Before the fix, a long-lived conversation's unbounded people list pushes the
+ * notification's parcel past the ~1 MB Binder limit and
+ * `NotificationManager.notify` throws [TransactionTooLargeException]
+ *
+ * Fails unless the attached people list is bounded.
+ */
+@RunWith(AndroidJUnit4::class)
+class BugleNotificationsTransactionSizeTest {
+
+    private lateinit var context: Context
+
+    @Before
+    fun setUp() {
+        context = ApplicationProvider.getApplicationContext()
+        val notificationManager = context.getSystemService(NotificationManager::class.java)
+
+        notificationManager.createNotificationChannel(
+            NotificationChannel(
+                NotificationChannelUtil.INCOMING_MESSAGES,
+                "Conversations",
+                NotificationManager.IMPORTANCE_HIGH,
+            ),
+        )
+
+        InstrumentationRegistry.getInstrumentation().uiAutomation.grantRuntimePermission(
+            context.packageName,
+            Manifest.permission.POST_NOTIFICATIONS,
+        )
+    }
+
+    @Test
+    fun notify_withManyAccumulatedMessages_doesNotExceedBinderLimit() {
+        val self = Person.Builder().setName("Me").build()
+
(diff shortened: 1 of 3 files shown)


m4pl (approved):

src/com/android/messaging/datamodel/BugleNotifications.java
```diff
@@ -314,6 +374,9 @@ private static void processAndSend(final MessageNotificationState state, final C
             style.setGroupConversation(conversation.mIsGroup);
             latestPerson = currentPerson;
         }
+        for (final Person person : limitPeople(peopleInConversation)) {
+            notifBuilder.addPerson(person);
+        }
```
inthewaves: `peopleInConversation` is coming from `reversedLineInfos` which appears to list messages from the oldest message to the newest (`mLineInfos` holds from newest to oldest: https://github.com/GrapheneOS/Messaging/blob/292881a75e709964b456e904883670488e11d38b/src/com/android/messaging/datamodel/MessageNotificationState.java#L290-L294)

`limitPeople` only takes at most the first 25 Persons in peopleInConversation. So with more than 25 distinct participants, the 25 oldest participants are retained and the most recent senders are dropped from `EXTRA_PEOPLE_LIST`

app/src/androidTest/java/com/android/messaging/datamodel/BugleNotificationsTransactionSizeTest.kt
```diff
@@ -0,0 +1,102 @@
...
+
+        val sender = Person.Builder()
+            .setName("Sender ${"0123456789".repeat(25)}")
+            .setKey("sender-1")
+            .build()
+
+        val style = MessagingStyle(self)
+        style.addMessage(MessagingStyle.Message("latest", 1L, sender))
+
+        val peopleInConversation = List(6000) { sender }
```
inthewaves: This seems to be making the same person for all 6000 senders, so the dedupe logic will just dedupe everything. We should add coverage for distinct `Person`s

app/src/test/kotlin/com/android/messaging/datamodel/BugleNotificationsMessageLimitTest.kt
```diff
@@ -0,0 +1,39 @@
...
+@RunWith(RobolectricTestRunner::class)
+class BugleNotificationsMessageLimitTest {
+
+    @Test
+    fun limitPeople_boundsAttachedPeople() {
+        val sender = Person.Builder().setName("Sender").setKey("sender-1").build()
+        val incoming = BugleNotifications.MAX_NOTIFICATION_PEOPLE * 20 + 7
+        val people = List(incoming) { sender }
+
+        val attached = BugleNotifications.limitPeople(people).size
```
inthewaves: This seems to be making the same person for all `BugleNotifications.MAX_NOTIFICATION_PEOPLE * 20 + 7` senders, so the dedupe logic will just dedupe everything. We should add coverage for distinct `Person`s

inthewaves (changes requested):

src/com/android/messaging/datamodel/BugleNotifications.java
```diff
@@ -314,6 +374,9 @@ private static void processAndSend(final MessageNotificationState state, final C
             style.setGroupConversation(conversation.mIsGroup);
             latestPerson = currentPerson;
         }
```
inthewaves: I think we should just move the limit logic directly into this function. e.g. `messageLineInfo.createPerson` is actually quite expensive as it makes actual Bitmaps, so we should be doing the dedupe and limiting in here. The per-Person work including Bitmaps is still $O(\text{line infos})$ instead of $O(\text{distinct Persons})$

Also, `List<Person> limitPeople` making a temporary list to only be iterated also seems inefficient, esp since `processAndSend` seems like a hot path (likely on every message in the convo)

Could use e.g `LinkedHashMap` with the Person stable key mapping to the person itself

inthewaves (changes requested):

src/com/android/messaging/datamodel/BugleNotifications.java
```diff
@@ -243,6 +255,74 @@ static String buildNotificationTag(final int type, final String conversationId)
...
+    private static Person getOrCreatePerson(
+            final Map<String, Person> peopleByKey,
+            final MessageNotificationState.MessageLineInfo lineInfo) {
+        final String key = personKey(lineInfo);
+        Person person = peopleByKey.get(key);
+        if (person == null) {
+            person = lineInfo.createPerson();
+            peopleByKey.put(key, person);
+        }
+        return person;
```
inthewaves: can be replaced with peopleByKey.computeIfAbsent

inthewaves (changes requested):

inthewaves (approved):

PR #143, opened 2026-06-07
Share Intent Screen: Java/XML to Kotlin/Compose migration

m4pl:
Migrated the Share Intent flow from Java + XML to Kotlin + Compose. Reused the composer building blocks from the Conversation migration: the compose bar, attachment preview cells, and subject chip were extracted into `common/components` so both screens share one implementation instead of duplicating it.

Reworked the share flow into an editable draft. Incoming text and attachments now land in a real composer with an attachment preview row, so the message can be edited, attachments removed, and the subject adjusted before anything is sent.

Added multi-target sending, so a single shared message can now be dispatched to several chats at once, plus a preview step that shows exactly what will be sent before confirming.

### Screenshots

| Before | After |
|--------|-------|
|Screenshot_20260607_134044|Screenshot_20260616_012002|
### Video
https://github.com/user-attachments/assets/9f124b14-4ea1-4f1a-b924-e930b9b3c91d


m4pl:
You can open the screen with the subject using:
```
adb shell "am start \
  -n com.android.messaging.debug/com.android.messaging.ui.conversationlist.ShareIntentActivity \
  -a android.intent.action.SEND -t text/plain \
  --es android.intent.extra.SUBJECT 'Test subject' \
  --es android.intent.extra.TEXT 'Hello from share'"
```

src/com/android/messaging/data/shareintent/repository/SharedAttachmentRepository.kt
```diff
@@ -0,0 +1,101 @@
...
+
+                else -> ConversationDraftAttachment(
+                    contentType = contentType,
+                    contentUri = persistedUri.toString(),
+                    displayName = displayName,
+                    durationMillis = durationMillis,
+                )
+            }
+        }
+    }
```
inthewaves, line 56: needs to restore behavior from old code + ideally add tests for those as well

https://github.com/GrapheneOS/Messaging/commit/9deefdf182d69f98d361d9d24de1e4892923dac5 for `ACTION_SEND` and  `ACTION_SEND_MULTIPLE` guards against file URIs

https://github.com/GrapheneOS/Messaging/commit/ce6fb7797b3362fd4574c095b4aa5d5d1d0de8a6 rejects private app data paths

src/com/android/messaging/ui/conversationpicker/host/share/ShareIntentActivity.kt
```diff
@@ -0,0 +1,102 @@
...
+
+        setContent {
+            AppTheme {
+                var isDraftLoading by remember { mutableStateOf(true) }
+                val conversationDraft by produceState<ConversationDraft?>(
+                    initialValue = null
+                ) {
+                    value = buildSharedConversationDraft(intent)
+                    isDraftLoading = false
+                }
```
inthewaves: Google Messages checks to see if `Activity.getInitialCaller` satisfies 

```kotlin
caller.checkContentUriPermission(
                      uri,
                      Intent.FLAG_GRANT_READ_URI_PERMISSION,
                  ) == PackageManager.PERMISSION_GRANTED
```

and rejects if that check fails, so we might as well also add that hardening too (would probably also naturally protect against what https://github.com/GrapheneOS/Messaging/commit/ce6fb7797b3362fd4574c095b4aa5d5d1d0de8a6 aims to fix)

src/com/android/messaging/domain/shareintent/usecase/BuildSharedConversationDraft.kt
```diff
@@ -0,0 +1,123 @@
...
+    suspend operator fun invoke(intent: Intent): ConversationDraft?
+}
+
+internal class BuildSharedConversationDraftImpl @Inject constructor(
+    private val resolveSharedContentType: ResolveSharedContentType,
+    private val sharedAttachmentRepository: SharedAttachmentRepository,
+) : BuildSharedConversationDraft {
+
+    override suspend fun invoke(intent: Intent): ConversationDraft? {
+        val subject = intent.getStringExtra(Intent.EXTRA_SUBJECT).orEmpty()
```
inthewaves: We can restore the subject fallback:

https://github.com/GrapheneOS/Messaging/blob/736203e47de8dfa01296e962a26dcdc80bc6ef22/src/com/android/messaging/ui/conversationlist/ShareIntentActivity.java#L85-L88

src/com/android/messaging/domain/shareintent/usecase/BuildSharedConversationDraft.kt
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    private suspend fun buildFromSendMultiple(
+        intent: Intent,
+        subject: String,
+    ): ConversationDraft? {
+        if (!ContentType.isImageType(intent.type)) {
+            return null
+        }
```
inthewaves: This looks like it will reject the entire `ACTION_SEND_MULTIPLE` if the top-level intent type is not image.

Compare with old code that checks the type per URI:

https://github.com/GrapheneOS/Messaging/blob/736203e47de8dfa01296e962a26dcdc80bc6ef22/src/com/android/messaging/ui/conversationlist/ShareIntentActivity.java#L128-L166

src/com/android/messaging/domain/shareintent/usecase/BuildSharedConversationDraft.kt
```diff
@@ -0,0 +1,123 @@
...
+        return draftOrNull(
+            messageText = intent.sharedMessageText(),
+            subject = subject,
+            attachments = attachments
+        )
+    }
+
+    private fun Intent.sharedMessageText(): String {
+        return getCharSequenceExtra(Intent.EXTRA_TEXT)?.toString().orEmpty()
+    }
```
inthewaves, line 196: Old code has a fallback to `getTextStringFromContentUri` if `Intent.EXTRA_TEXT` was null. `intent.sharedMessageText()` only queries `Intent.EXTRA_TEXT`

https://github.com/GrapheneOS/Messaging/blob/736203e47de8dfa01296e962a26dcdc80bc6ef22/src/com/android/messaging/ui/conversationlist/ShareIntentActivity.java#L104-L114

Keep in mind https://github.com/GrapheneOS/Messaging/commit/ce6fb7797b3362fd4574c095b4aa5d5d1d0de8a6 from old code rejects private app data paths

src/com/android/messaging/domain/conversationpicker/usecase/SendContentToConversations.kt
```diff
@@ -0,0 +1,92 @@
...
+
+    private suspend fun sendToConversation(
+        conversationId: String,
+        draft: ConversationDraft,
+    ) {
+        try {
+            sendConversationDraft(conversationId, draft).collect()
+        } catch (exception: SendConversationDraftException) {
+            LogUtil.w(TAG, "Failed to send shared draft to $conversationId", exception)
+        }
```
inthewaves, line 105: this should probably be surfaced to the user instead of silent logged failure

src/com/android/messaging/ui/conversationpicker/ConversationPickerScreen.kt
```diff
@@ -0,0 +1,531 @@
...
+    }
+
+    BackHandler(enabled = !uiState.draft.isReviewing && uiState.targets.isSearchActive) {
+        searchState.clearText()
+        onAction(Action.SearchClosed)
+    }
+}
+
+@Composable
+private fun PickerScaffold(
```
inthewaves, line 208: note that this doesn't have the New message button like old code, so users cannot share to an arbitrary number/email that is not already a conversation or visible contact

src/com/android/messaging/ui/conversationpicker/ConversationPickerScreen.kt
```diff
@@ -0,0 +1,531 @@
...
+        },
+    ) { contentPadding ->
+        Box(
+            modifier = Modifier
+                .fillMaxSize()
+                .padding(top = contentPadding.calculateTopPadding())
+                .clip(MaterialTheme.contentSurfaceShape)
+                .background(MaterialTheme.colorScheme.background),
+        ) {
+            MessageComposeBar(
```
inthewaves: I think there is too much empty space here



There's also a parity difference with normal conversation attachment sending, e.g. when sending from inside conversation > Press plus button > Choose media, after media is picked, there is a full screen preview (scrollable if multiple media picked). Here, tapping the small images doesn't bring up preview

inthewaves (changes requested):

src/com/android/messaging/ui/conversationpicker/ConversationPickerScreen.kt
```diff
@@ -0,0 +1,531 @@
...
+                onRemove = { onAction(Action.SelectionToggled(it)) },
+                onProceed = { onAction(Action.ProceedToReviewClicked) },
+                showProceedButton = true,
+            )
+        }
+    }
+}
+
+@Composable
+private fun PickerReviewScaffold(
```
inthewaves, line 444: for multi-SIM environment, it should be clear in here what SIM will be used for the send and add way for user to choose. For single tap for 1 sender, that opens up the conversation as a draft; it's probably okay in that case, since the SIM is shown there anyway

src/com/android/messaging/ui/conversationpicker/mapper/ContactUiStateMapper.kt
```diff
@@ -0,0 +1,42 @@
...
+    override fun map(
+        contacts: ImmutableList<Contact>,
+    ): ImmutableList<TargetUiState.Contact> {
+        return contacts
+            .mapNotNull(::toContactUiState)
+            .toImmutableList()
+    }
+
+    private fun toContactUiState(contact: Contact): TargetUiState.Contact? {
+        val destination = contact.destinations.firstOrNull() ?: return null
```
inthewaves: If a contact has multiple phone numbers then this will only show the first phone number. In fact, when searching for the contact's second, 3rd, etc. phone number, it will only show result for the first phone number.

Generally this new `ConversationPicker` is missing features from the `NewChatScreen` / `RecipientSelectionContent` such as listing multiple numbers + the type of each number for each contact. See the other comment about this duplicating a lot of existing contact picker code

src/com/android/messaging/ui/conversationpicker/ConversationPickerScreen.kt
```diff
@@ -0,0 +1,531 @@
...
+
+private fun isReadContactsPermissionGranted(context: Context): Boolean {
+    return ContextCompat.checkSelfPermission(
+        context,
+        Manifest.permission.READ_CONTACTS,
+    ) == PackageManager.PERMISSION_GRANTED
+}
+
+@Composable
+private fun PickerContent(
```
inthewaves, line 143: This adds a second contact/recipient picker path beside the existing new chat recipient picker in `NewChatScreen`, so the app will now have two contact/recipient picker implementations

`TargetsDelegate` now duplicates contact permission handling, debounced contact search, pagination, page merging, and contact row mapping

`ConversationPickerViewModel` duplicates contact-to-conversation resolution. That creates functional drift as this picker only exposes the first destination for a contact, lacks typed-number recipients, and has separate pagination behavior (https://github.com/GrapheneOS/Messaging/pull/143#discussion_r3393494546)

For all contacts display, I think it would be better for the share flow to reuse/adapt `RecipientSelectionContent` and `RecipientPickerDelegate` from the new chat screen instead of maintaining a parallel contact picker. Of course, the "Recent conversations" would be a new addition

src/com/android/messaging/ui/conversationpicker/TargetLoadMore.kt
```diff
@@ -0,0 +1,41 @@
...
+
+            totalItemsCount > 0 && lastVisibleIndex >= prefetchThreshold
+        }
+            .distinctUntilChanged()
+            .collect { isNearEnd ->
+                if (isNearEnd) {
+                    currentOnLoadMore()
+                }
+            }
+    }
```
inthewaves: This is almost a duplicate of https://github.com/GrapheneOS/Messaging/blob/e0f99a68ca785bda6a53ce8f55bf3112a6945c6e/src/com/android/messaging/ui/conversation/recipientpicker/component/RecipientSelectionContactsContent.kt#L241-L283, except `RecipientSelectionContactsContent` keys the effect on more keys, etc.

src/com/android/messaging/ui/conversationpicker/ConversationPickerScreen.kt
```diff
@@ -0,0 +1,531 @@
...
+        enabled = !uiState.draft.isReviewing &&
+            uiState.targets.selection.selectedIds.isNotEmpty(),
+    ) {
+        onAction(Action.SelectionCleared)
+    }
+
+    BackHandler(enabled = !uiState.draft.isReviewing && uiState.targets.isSearchActive) {
+        searchState.clearText()
+        onAction(Action.SearchClosed)
+    }
```
inthewaves: Inconsistency between system back button and toolbar back button

1. Open share picker
2. Tap search and type a query
3. Long-press a result to select it
4. Toolbar shows selected count, not the search field.
5. Press toolbar nav back button and see selection clears.
6. Repeat and press system backbutton. Then search closes/clears first, while selection remains.

inthewaves (changes requested):

src/com/android/messaging/domain/shareintent/usecase/ResolveSharedContentType.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    private fun extractFromMetadata(
+        uri: Uri,
+        fallbackType: String?,
+    ): String? {
+        val retriever = MediaMetadataRetrieverWrapper()
+
+        return try {
+            retriever.setDataSource(uri)
+            retriever.extractMetadata(MediaMetadataRetriever.METADATA_KEY_MIMETYPE) ?: fallbackType
```
inthewaves: These open asset fds and get invoked in `BuildSharedConversationDraftImpl` without ensuring coroutine context is on an IO dispatcher

src/com/android/messaging/ui/conversationpicker/host/share/ShareIntentActivity.kt
```diff
@@ -0,0 +1,102 @@
...
+
+        enableEdgeToEdge()
+
+        setContent {
+            AppTheme {
+                var isDraftLoading by remember { mutableStateOf(true) }
+                val conversationDraft by produceState<ConversationDraft?>(
+                    initialValue = null
+                ) {
+                    value = buildSharedConversationDraft(intent)
```
inthewaves: `produceState` uses `LaunchedEffect(Unit)` which runs on main thread by default (this part is probably fine, just need to make IO paths main-safe)

src/com/android/messaging/domain/shareintent/usecase/BuildSharedConversationDraft.kt
```diff
@@ -0,0 +1,123 @@
...
+) : BuildSharedConversationDraft {
+
+    override suspend fun invoke(intent: Intent): ConversationDraft? {
+        val subject = intent.getStringExtra(Intent.EXTRA_SUBJECT).orEmpty()
+
+        return when (intent.action) {
+            Intent.ACTION_SEND -> buildFromSend(intent, subject)
+            Intent.ACTION_SEND_MULTIPLE -> buildFromSendMultiple(intent, subject)
+            else -> null
+        }
```
inthewaves: These can lead to opening asset fds from `resolveSharedContentType`. We can just make `BuildSharedConversationDraftImpl` main-safe (https://developer.android.com/kotlin/coroutines/coroutines-best-practices#main-safe) by adding coroutine context switch here (or making  `ResolveSharedContentType` suspending and run both `ContentResolver.getType()` and metadata extraction under the injected `IoDispatcher`)

inthewaves (changes requested):
Potential IO on main thread

m4pl:
@inthewaves I've changed my approach to the screens and moved many components into shared modules so they can be reused. I also fixed the code based on your comments.

https://github.com/user-attachments/assets/ed4b8da9-2f43-4bdc-a0d5-5f459d17d93e

src/com/android/messaging/domain/shareintent/usecase/BuildSharedConversationDraft.kt
```diff
@@ -0,0 +1,209 @@
...
+            attachments = attachments.toImmutableList(),
+        )
+    }
+
+    private suspend fun persistAttachment(
+        sourceUri: Uri,
+        contentType: String?,
+        caller: ComponentCaller,
+    ): ConversationDraftAttachment? {
+        if (contentType.isNullOrBlank()) {
```
RankoR, line 218: Could be the first branch of `return when`

src/com/android/messaging/ui/conversationpicker/delegate/TargetsDelegate.kt
```diff
@@ -0,0 +1,240 @@
...
+    val state: StateFlow<TargetsUiState>
+    val selectedIds: Flow<ImmutableSet<String>>
+    fun bind(scope: CoroutineScope)
+    fun setSearchActive(active: Boolean)
+    fun setSearchQuery(query: String)
+    fun toggleSelection(target: TargetUiState)
+    fun clearSelection()
+    fun loadMoreRecent()
+    fun collapseRecent()
+    fun currentSelectedTargets(): ImmutableList<TargetUiState>
```
RankoR: Could be `val` with `get()`

src/com/android/messaging/ui/conversationpicker/formatter/TargetTextFormatter.kt
```diff
@@ -0,0 +1,23 @@
...
+
+    fun wrap(text: String): String {
+        return BidiFormatter.getInstance().unicodeWrap(text, TextDirectionHeuristicsCompat.LTR)
+    }
+
+    fun detailsOrNull(
+        name: String,
+        value: String?,
+    ): String? {
+        val details = value?.takeIf {
```
RankoR:
```
return value
   ?.takeIf {
      it.isNotEmpty() && it != name
   }
   ?.let(::wrap)
```

is easier to read imo

src/com/android/messaging/domain/shareintent/usecase/BuildSharedConversationDraft.kt
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    private fun ComponentCaller.canReadContent(uri: Uri): Boolean {
+        return runCatching {
+            checkContentUriPermission(
+                uri,
+                Intent.FLAG_GRANT_READ_URI_PERMISSION,
+            ) == PackageManager.PERMISSION_GRANTED
+        }.getOrDefault(false)
+    }
```
inthewaves: 1. We can add `UriUtil.isFileUri(uri)` and `FileUtil.isInPrivateDir(uri)` checks here. I looked more closely at `FileUtil.isInPrivateDir(uri)` and it seems to do more than what the method is named for, i.e. it rejects fd-backed paths too, so the caller check alone is probably not sufficient
2. I tried using https://github.com/GrapheneOS/Gallery/pull/2 to share an image and it seems to be failing with `Ignoring shared attachment without caller read permission` when sharing an image from here. However, Google Photos sharing the same image had no issue.

    This appears to be because ReFra passes `content://media/external/images/media/<id>` straight through; Google Photos shares through its own provider/grant infrastructure. But I also used the ReFra fork to share the same image to Google Messages which Google Messages also accepted. I looked more closely at the Google Messages code and this caller check is behind a feature flag that appears to be off right now

So I think we can remove the `checkContentUriPermission` check for now but add the  `UriUtil.isFileUri(uri)` and `FileUtil.isInPrivateDir(uri)` checks

src/com/android/messaging/domain/shareintent/usecase/BuildSharedConversationDraft.kt
```diff
@@ -0,0 +1,209 @@
...
+        intent: Intent,
+        subject: String,
+        caller: ComponentCaller,
+    ): ConversationDraft? {
+        val contentUri = IntentCompat.getParcelableExtra(
+            intent,
+            Intent.EXTRA_STREAM,
+            Uri::class.java,
+        )
+        val contentType = resolveSharedContentType(contentUri, intent.type)
```
inthewaves, line 56: We should do security checks on the uri here instead of in `persistAttachment`/`sharedTextFromContent` to avoid parsing URIs that shouldn't be processed. `resolveSharedContentType` can potentially open the URI

Note URI could be invalid but other stuff such as text/subject could still be valid. Not sure if it's worth rejecting the entire thing if the attachment goes missing due to failing URI checks, but either way the user should also be informed it happening

src/com/android/messaging/domain/shareintent/usecase/BuildSharedConversationDraft.kt
```diff
@@ -0,0 +1,209 @@
...
+            Intent.EXTRA_STREAM,
+            Uri::class.java,
+        ).orEmpty()
+
+        val attachments = mutableListOf<ConversationDraftAttachment>()
+        val sharedTexts = mutableListOf<String>()
+
+        sharedUris.forEach { uri ->
+            val contentType = resolveSharedContentType(uri, intent.type)
+            when {
```
inthewaves: We should do security checks on each uri here instead of in `persistAttachment`/`sharedTextFromContent`  to avoid parsing URIs that shouldn't be processed.  `resolveSharedContentType` can potentially open the URI

Note URI could be invalid but other stuff such as text/subject could still be valid. Not sure if it's worth rejecting the entire thing if the attachment goes missing due to failing URI checks, but either way the user should also be informed it happening

src/com/android/messaging/ui/conversationpicker/host/share/ShareIntentActivity.kt
```diff
@@ -0,0 +1,109 @@
...
+
+        setContent {
+            AppTheme {
+                var isDraftLoading by remember { mutableStateOf(true) }
+                val conversationDraft by produceState<ConversationDraft?>(
+                    initialValue = null,
+                ) {
+                    value = buildSharedConversationDraft(intent, initialCaller)
+                    isDraftLoading = false
+                }
```
inthewaves: There should be some user-visible error handling when no draft is available (e.g. all URIs are invalid URIs rejected by security checks). Currently when all URIs are invalid / draft is null, the picker still allows user to pick conversations but going to them will have no attachments or content (assuming only an image was shared with no subject, etc.). We should probably also have user-facing messages when any URI fails or errors occur

inthewaves (changes requested):

src/com/android/messaging/domain/shareintent/usecase/BuildSharedConversationDraft.kt
m4pl: I'd remove `checkContentUriPermission` for now and add it to the backlog. As I understand it the fix belongs on the sender side, getting the grant right, and then we can bring the caller check back.

On the fd path in `isInPrivateDir`, the uid model already covers us as far as I can tell. A provider can only return a descriptor to a file it can open as itself, so it can't reach our data or another app's. It also rejects content URIs backed by a file under `/data`, like Chrome's share cache, so it blocks valid shares.

inthewaves (approved):
LGTM, looks clean now

PR #147, opened 2026-06-09
Apply SIM picked in new chat screen when sending

### LINKED ISSUE #142: New SIM picker in start new conversation screen doesn't appear to work
For existing conversations, it seems to still use the old SIM for that conversation

For new conversations (can also delete a conversation then start new conversation for that same phone number), it will actually use the default sending SIM from the SIM settings in the OS

The UI will can also be misleading for new conversations (or conversations that were deleted before). It can say it will send as A, but if the default SMS SIM is B, then messages will actually get sent as B. If you then change the default SMS SIM in OS Settings to A, then it will send as A. If you change it back to B, it will send as B but the UI will still say A. The UI only becomes consistent for new conversations when a SIM is selected in the conversation SIM picker

RankoR:
Fixes https://github.com/GrapheneOS/Messaging/issues/142


inthewaves (approved):
Fixes the in-app start new conversation picker

PR #153, opened 2026-06-12
Convert app class to Kotlin

RankoR:
Just a nice-to-have to simplify future fixes like https://github.com/GrapheneOS/Messaging/issues/97 (fix is WIP)


inthewaves (approved):

src/com/android/messaging/BugleApplication.kt
```diff
@@ -0,0 +1,270 @@
...
+
+    // Called from thread started in FactoryImpl.register(), not in tests.
+    fun initializeAsync(factory: Factory) {
+        traceSection(sectionName = "app.initializeAsync") {
+            maybeHandleSharedPrefsUpgrade(factory = factory)
+            MmsConfig.load()
+        }
+    }
+
+    override fun onLowMemory() {
```
m4pl, line 115: `onLowMemory()` is deprecated and no longer called on recent Android versions. We should move this cleanup to `onTrimMemory()` instead.

m4pl (approved):

src/com/android/messaging/BugleApplication.kt
RankoR: This class needs a rewrite later, to migrate to modern architecture. So decided to keep everything as-is for now.

PR #154, opened 2026-06-13
Change hardcoded colors to more neutral, and add some basic color schemes

ddagunts:
Hello, first off, a big thanks for the GrapheneOS project.  I am hoping you will consider this PR, or consider a similar/better implementation.  The default color scheme in the Messaging app for incoming messages has white text on green background.  Green and white is a fairly common color blindness - for me it is difficult to read incoming messages.  There are other messaging and SMS apps available, but unfortunately I haven't found an SMS app that works reliably in secondary profiles (well known issues).  

This changes the old green defaults to grays, and provides a few color schemes which, while not designer quality, do seem to be easier to read.

Thank you!

### DIFF
--- a/res/values/colors.xml
+++ b/res/values/colors.xml
@@ -15,12 +15,12 @@
     limitations under the License.
 -->
 <resources>
-    <color name="primary_color">#689F38</color>
+    <color name="primary_color">#616161</color>
 
     <!--  Action bar -->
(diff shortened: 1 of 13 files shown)


ddagunts:
I can see the code is in a state of modernization, so it's understandable if this change is undesirable and declined.  I would ask, in such a case, to consider changing the hue on the message bubbles to a darker green.

Thank you.

PR #156, opened 2026-06-15
Migrate legacy conversation notification settings to channels

RankoR:
- Migrate legacy per-conversation notification preferences into Android conversation notification channels once at app startup.
- Preserve existing per-conversation mute, ringtone, and vibration settings during migration without overwriting already-created channels.
- Stop runtime conversation-channel creation from reading legacy database notification fields, so new channels inherit the parent incoming-message channel defaults.
- Open conversation notification settings through the parent incoming-message channel plus `EXTRA_CONVERSATION_ID`.
- Add focused Robolectric coverage for the migration, channel creation behavior, and conversation-settings intent handling.

Fixes https://github.com/GrapheneOS/Messaging/issues/97


src/com/android/messaging/domain/notification/usecase/MigrateConversationNotificationChannels.kt
```diff
@@ -0,0 +1,180 @@
...
+
+    override operator fun invoke(): Boolean {
+        val prefs = BuglePrefs.getApplicationPrefs()
+        if (isMigrationComplete(prefs = prefs)) {
+            return false
+        }
+
+        return runBlocking {
+            migrateIfNeeded(prefs = prefs)
+        }
```
inthewaves, line 38: technically blocks main thread as this runs in application, but I suppose there's no better way to manage app init right now, especially if this needs to be run before other db code

inthewaves (approved):

PR #157, opened 2026-06-18
Add issue templates

RankoR:


m4pl (approved):
The forms look good overall. Could we add a required checkbox confirming that the reporter searched for existing issues to reduce duplicates? It would also be helpful to request the device model and build number, since the app version alone may not be enough to reproduce a problem. Do we want blank issues to be available to everyone, or should `blank_issues_enabled` be set to `false` so they remain available only to maintainers?

RankoR:
@m4pl Not sure about a checkbox, I wanted to simplify the form as much as possible to not scare off a user. 

As for the `blank_issues_enabled` - yes, ideally we shouldn't allow blank issues creation, but as I understand, setting it to `false` will also disable them for us (write access required to create them, we don't have it).

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,36 @@
...
+  - type: textarea
+    id: logs
+    attributes:
+      label: Logcat / stack trace
+      description: >
+        Paste relevant logcat output here, or drag and drop a log file into
+        this field to attach it. ⚠️ This is a messaging app — please scrub
+        phone numbers, contact names, and message bodies first. If the logs
+        contain sensitive data you'd rather not post publicly, email them to
+        logs@grapheneos.org instead and mention that here.
```
inthewaves: We could probably add GrapheneOS instructions on how to get logs here

e.g. Long pressing the Messages app in Home or launcher drawer and going to app info, or OS Settings > Apps > See all apps > Search for Messaging. Then View logs in app info

Same for the system log, system log requires being on the Owner user can accessed from OS Settings > System > View logs

inthewaves (changes requested):

inthewaves (approved):

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,46 @@
+name: Bug report
+description: Report a problem with the Messaging app
+type: Bug
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for filing a bug. Please fill in as much as you can
```
soupslurpr: Very minor but there should probably be a period here at the end.

soupslurpr: I think it could be tweaked a bit, "filing a bug" sounds a bit off and there could be some more detail in general. Maybe something like the following? Thoughts?

"Thanks for reporting this! The more detail you can give us, the faster we can help. If you're short on time, clear reproduction steps and what you expected vs. what happened are the most useful."

soupslurpr (changes requested):

.github/ISSUE_TEMPLATE/bug_report.yml
RankoR: Done

soupslurpr (approved):

PR #160, opened 2026-06-23
Rewrite Class 0 dialog with Compose

RankoR:
image
image


res/values/styles.xml
```diff
@@ -60,6 +60,11 @@
         <item name="android:background">@android:color/white</item>
     </style>
 
+    <style name="BugleTheme.ClassZeroActivity" parent="@style/BugleTheme.DialogActivity">
```
m4pl, line 72: Can we clarify why `ClassZeroActivity` needs its own theme? If it is only for the transparent Compose dialog host, maybe all related window config should live here

src/com/android/messaging/ui/classzero/ClassZeroScreen.kt
```diff
@@ -0,0 +1,237 @@
...
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.tooling.preview.PreviewLightDark
+import androidx.compose.ui.unit.dp
+import com.android.messaging.R
+import com.android.messaging.ui.classzero.model.ClassZeroUiState
+import com.android.messaging.ui.core.MessagingPreviewTheme
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+internal fun ClassZeroScreen(
```
m4pl, line 40: In landscape, the dialog can overflow because only the message body scrolls. Can we make the whole dialog content scrollable?

src/com/android/messaging/ui/classzero/ClassZeroScreen.kt
```diff
@@ -0,0 +1,237 @@
...
+    uiState: ClassZeroUiState,
+    onSaveClicked: () -> Unit,
+    onCancelClicked: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    BasicAlertDialog(
+        modifier = modifier
+            .widthIn(max = 360.dp),
```
m4pl: Why is the dialog capped at 360.dp? Can we use the default Material dialog width, or move this to a named constant with context?

src/com/android/messaging/ui/classzero/ClassZeroActivity.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    @Inject
+    @MainDispatcher
+    lateinit var mainDispatcher: CoroutineDispatcher
+
+    private val screenModel: ClassZeroViewModel by viewModels()
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        requestWindowFeature(Window.FEATURE_NO_TITLE)
```
m4pl: Can these window settings live in `BugleTheme.ClassZeroActivity` instead?

m4pl (approved):

thestinger:
GitHub says there are conflicts, can you rebase it?

src/com/android/messaging/ui/ClassZeroActivity.java
```diff
@@ -1,205 +0,0 @@
...
-
-    @Override
-    protected void onStop() {
-        super.onStop();
-        mHandler.removeMessages(ON_AUTO_SAVE);
-        if (VERBOSE) {
-            Log.d(TAG, "onStop time = " + Long.toString(mTimerSet)
-                    + " " + this.toString());
-        }
-    }
```
inthewaves, line 186: Note that old ClassZeroActivity stopped the timer (the `ON_AUTO_SAVE` message) in `onStop`. In `onStart`, it'll send the `ON_AUTO_SAVE` immediately if it was past the time already

src/com/android/messaging/ui/classzero/ClassZeroViewModel.kt
```diff
@@ -0,0 +1,257 @@
...
+            withContext(mainDispatcher) {
+                finishCurrentMessage()
+            }
+        }
+    }
+
+    private fun cancelTimeout() {
+        timeoutJob?.cancel()
+        timeoutJob = null
+    }
```
inthewaves, line 203: Note that unlike the old ClassZeroActivity, the timer is not cancelled when the activity is stopped. This means other messages in the queue will be processed and continued when e.g. user backgrounds the app with the class zero dialog visible. The user may not get a chance to see other messages in the queue if it keeps processing while the activity is stopped

inthewaves (changes requested):

inthewaves:
CI build failures:

```
> Task :app:minifyReleaseWithR8

e: file:///home/runner/work/Messaging/Messaging/app/src/test/kotlin/com/android/messaging/ui/ActivityPermissionGateTest.kt:25:40 Cannot infer type for type parameter 'T'. Specify it explicitly.
e: file:///home/runner/work/Messaging/Messaging/app/src/test/kotlin/com/android/messaging/ui/ActivityPermissionGateTest.kt:25:40 Inapplicable candidate(s): fun <T> setOf(vararg elements: T): Set<T>
> Task :app:compileDebugUnitTestKotlin FAILED
e: file:///home/runner/work/Messaging/Messaging/app/src/test/kotlin/com/android/messaging/ui/ActivityPermissionGateTest.kt:31:9 Unresolved reference 'ClassZeroActivity'.
e: file:///home/runner/work/Messaging/Messaging/app/src/test/kotlin/com/android/messaging/ui/ActivityPermissionGateTest.kt:31:34 Cannot infer type for type parameter 'T'. Specify it explicitly.
e: file:///home/runner/work/Messaging/Messaging/app/src/test/kotlin/com/android/messaging/ui/ActivityPermissionGateTest.kt:31:34 Unresolved reference. None of the following candidates is applicable because of a receiver type mismatch:
val <T> KClass<T>.java: Class<T>
```

inthewaves (approved):

PR #162, opened 2026-06-25
Conversation List Screen: Java/XML to Kotlin/Compose migration

m4pl:
Migrated the Conversation List screen from Java + XML to Kotlin + Compose. Please verify the list animations (archive, undo, pin/unpin, scroll-to-top) manually on a device, as I had some issues after writing the code. I also tried to move some components to ui/common/components.

### Screenshots

| Before | After |
|--------|-------|
|Screenshot_20260625_021402|Screenshot_20260625_021447|

### Video
https://github.com/user-attachments/assets/b04feb46-bb36-4c3f-b861-63cda3d58a4a


src/com/android/messaging/ui/conversationlist/model/ConversationListItemUiModel.kt
```diff
@@ -0,0 +1,68 @@
+package com.android.messaging.ui.conversationlist.model
+
+import androidx.compose.runtime.Immutable
+import com.android.messaging.data.conversationlist.model.ConversationListMessageStatus
+
+@Immutable
+internal data class ConversationListItemUiModel(
```
inthewaves, line 8: If the last message was an incoming MMS message that hasn't been downloaded, not enough info is displayed to the user:



Google Messages for reference on the same message:

src/com/android/messaging/ui/conversationlist/ui/item/SwipeableConversationListItem.kt
```diff
@@ -0,0 +1,470 @@
...
+
+    val icon = when {
+        isArchive -> Icons.Filled.Archive
+        else -> Icons.Filled.MarkChatUnread
+    }
+
+    val description = when {
+        isArchive -> stringResource(R.string.action_archive)
+        else -> stringResource(R.string.mark_as_unread)
+    }
```
inthewaves, line 464: This seems like it'll use `R.string.mark_as_unread` for content description even if the swipe action is mark as read

src/com/android/messaging/datamodel/DatabaseUpgradeHelper.java
```diff
@@ -63,6 +66,13 @@ private int upgradeToVersion2(final SQLiteDatabase db) {
         return 2;
     }
 
+    private int upgradeToVersion3(final SQLiteDatabase db) {
+        db.execSQL("ALTER TABLE " + DatabaseHelper.CONVERSATIONS_TABLE + " ADD COLUMN " +
+                DatabaseHelper.ConversationColumns.PINNED + " INT DEFAULT(0)");
+        LogUtil.i(TAG, "Upgraded database to version 3");
+        return 3;
+    }
```
inthewaves: Seems to be missing the index creation

src/com/android/messaging/ui/conversationlist/ConversationListActivity.kt
```diff
@@ -0,0 +1,44 @@
...
+        super.onCreate(savedInstanceState)
+
+        if (isFinishing) {
+            return
+        }
+
+        enableEdgeToEdge()
+
+        setContent {
+            AppTheme {
```
inthewaves, line 28: The theme for `ConversationListActivity` in the manifest is still `@style/LaunchTheme`, maybe we should use `Theme.Compose` instead

src/com/android/messaging/ui/conversationlist/ui/ConversationListScreen.kt
```diff
@@ -0,0 +1,603 @@
...
+    val pinAnimationController = rememberOverlayReorderAnimationController(
+        key = ConversationListItemUiModel::conversationId,
+        isSettled = { item, anchorToTop -> item.isPinned == anchorToTop },
+    )
+
+    var pendingAddContactDestination by rememberSaveable { mutableStateOf<String?>(null) }
+    var pendingDelete by rememberSaveable { mutableStateOf(false) }
+    var pendingBlockConversationId by rememberSaveable { mutableStateOf<String?>(null) }
+    var pendingBlockDestination by rememberSaveable { mutableStateOf<String?>(null) }
+    var pendingSnooze by rememberSaveable { mutableStateOf(false) }
```
inthewaves: These use `rememberSaveable` but they seem like they can be out of sync with the actual state in the ViewModel, e.g

1. Enable Android developer option "Don't keep activities", or otherwise force the app process/activity to be destroyed while in the background
2. Open Messaging, long-press a conversation, and tap the delete action so the confirmation dialog is visible.
3. Background the app, then return to Messaging from Recents.
4. Observe that the delete dialog can be restored even though the new view-model state no longer has the original selection. When only one conversation was selected for deletion, the dialog title can change from "Delete this conversation" to "Delete these conversations?" after the process death, and the delete button is a no-op.
5. Repeat with snooze, the dialog will be a no-op

It seems that for block or add contact from a row action, the restored dialog can still reference the old destination after the underlying list state has been recreated

It might be better to just make these `remember`

src/com/android/messaging/ui/conversationlist/ui/ConversationListDialogs.kt
```diff
@@ -0,0 +1,195 @@
...
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.PreviewLightDark
+import com.android.messaging.R
+import com.android.messaging.ui.common.components.SnoozeChatDialog
+import com.android.messaging.ui.conversationlist.model.ConversationListAction as Action
+import com.android.messaging.ui.core.MessagingPreviewTheme
+
+@Composable
+internal fun ConversationListDialogs(
+    selectedCount: Int,
```
inthewaves, line 18: Maybe we should sanity check this against a selectedCount of 0. Dialogs are still shown when selectedCount is 0 it seems (e.g. the Don't keep activities enabled with Delete and Snooze dialogs)

src/com/android/messaging/ui/conversationlist/ui/ConversationListDialogs.kt
```diff
@@ -0,0 +1,195 @@
...
+
+    if (isSnoozeVisible) {
+        SnoozeChatDialog(
+            onDismiss = onDismissSnooze,
+            onConfirm = { option ->
+                onAction(Action.SnoozeOptionSelected(option))
+                onDismissSnooze()
+            },
+        )
+    }
```
inthewaves: The snooze chat dialog should use plural strings when multiple chats are selected. The dialog currently says "Snooze chat" and "You won't get notifications from this chat for the selected period"

src/com/android/messaging/ui/conversationlist/ui/item/ConversationListItemRow.kt
```diff
@@ -0,0 +1,514 @@
...
+        ),
+        onLongClick = onLongClick,
+        color = itemContainerColor(item),
+        subtitleContent = {
+            ConversationListItemBody(item)
+        },
+        trailingContent = {
+            ConversationListItemTrailing(item)
+        },
+    )
```
inthewaves, line 80: This appears to regress a lot of the accessibility content descriptions compared to the old Messages code, e.g. 

https://github.com/GrapheneOS/Messaging/blob/12/src/com/android/messaging/ui/conversationlist/ConversationListItemView.java#L304

https://github.com/GrapheneOS/Messaging/blob/12/res/values/strings.xml#L372-L389

When I use TalkBack with the old messages app, for successful messages, it will say stuff like 

> Message from `<contact>`, `<message body>`, time `<time>`

This is from 

```xml
<string name="one_on_one_incoming_successful_message_prefix">Message from <xliff:g id="sender">%1$s</xliff:g>: <xliff:g id="message">%2$s</xliff:g>. Time: <xliff:g id="time">%3$s</xliff:g>.</string>
```

e.g.

> Message from Bob, "Hello there". Time: 33 minutes

and it will not select the avatar. The only other things TalkBack will scroll over are attachments, e.g. for an audio attachment, it will say "Play audio attachment"

When I use TalkBack with this PR, it will just say, 

> Bob, 33 minutes, "Hello there"

and for the top conversation it might say "in list, 57 items"

Swiping on TalkBack to try to go to the next may select the avatar (when the contact doesn't have a custom avatar in contacts) and it'll just say "capital F" for a contact with first name F, or "unlabeled" for just a number or a group avatar

src/com/android/messaging/ui/conversationlist/ui/item/ConversationListItemRow.kt
```diff
@@ -0,0 +1,514 @@
...
+
+@Composable
+private fun ConversationListItemBadgeIcon(icon: ImageVector) {
+    Icon(
+        imageVector = icon,
+        contentDescription = null,
+        modifier = Modifier.size(ItemBadgeIconSize),
+        tint = MaterialTheme.colorScheme.onSurfaceVariant,
+    )
+}
```
inthewaves, line 188: We should avoid `null` `contentDescription`s and describe what the badge is indicating, or the accessibility hint on the conversation list item itself should mention the status instead of the badge icon being traversable

inthewaves (changes requested):

src/com/android/messaging/datamodel/DatabaseUpgradeHelper.java
inthewaves: As a side note, we should probably eventually add tests to ensure a migrated database is the same as newly created database

src/com/android/messaging/ui/conversationlist/ui/ConversationListScreen.kt
```diff
@@ -0,0 +1,603 @@
...
+            modifier = Modifier
+                .fillMaxSize()
+                .padding(top = contentPadding.calculateTopPadding())
+                .onGloballyPositioned { coordinates ->
+                    pinAnimationController?.updateContentTop(coordinates.boundsInRoot().top)
+                }
+                .background(backdropColor)
+                .clip(ContentCornerShape)
+                .background(MaterialTheme.colorScheme.background),
+        ) {
```
inthewaves, line 420: Needs to be inset aware when using 3-button gesture navigation; the camera cutout can also clip into the sides especially in the other landscape orientation (camera hole not shown)

(The appbar itself seems to be inset aware)

src/com/android/messaging/ui/conversationlist/ui/ConversationListItemRow.kt
```diff
@@ -0,0 +1,514 @@
...
+@Composable
+private fun ConversationListItemTrailing(item: ConversationListItemUiModel) {
+    Row(
+        modifier = Modifier.padding(end = ItemTrailingEndPadding),
+        horizontalArrangement = Arrangement.spacedBy(ItemHeaderSpacing),
+        verticalAlignment = Alignment.CenterVertically,
+    ) {
+        ConversationListItemPreviewThumbnail(item.snippet.preview)
+    }
+}
```
inthewaves: I think the thumbnail placement makes it look unaligned or not uniform, because it offsets the timestamp compared to other conversations that don't have the thumbnail

inthewaves (changes requested):

src/com/android/messaging/ui/conversationlist/model/ConversationListItemUiModel.kt
m4pl: Added more information:
image

src/com/android/messaging/ui/conversationlist/ui/ConversationListItemRow.kt
m4pl: I decided to add a preview before the message text. What do you think, @inthewaves?
Screenshot_20260630_163052

inthewaves: Looks good!

src/com/android/messaging/ui/conversationlist/ui/item/ConversationListItemContentDescription.kt
```diff
@@ -0,0 +1,227 @@
...
+): String {
+    val message = conversationListItemContentDescriptionMessage(item)
+    val time = remember(item.timestampMillis) {
+        Dates.getConversationTimeString(item.timestampMillis).toString()
+    }
+    val spec = conversationListItemContentDescriptionSpec(
+        item = item,
+        message = message,
+        time = time,
+    )
```
inthewaves, line 24: For days of the week, TalkBack will say the shortened word instead of the full word

e.g. if the date just says "Fri", TalkBack should say Friday instead of Fri ("fry")

inthewaves (changes requested):

m4pl:
Two fixes:
- Archived chats coming back on their own. A sync was resetting `ARCHIVE_STATUS` because `keepArchived` only checked the sync cache, not the user's archive state in the DB (old AOSP behavior, now visible with the optimistic list). Fix: also check the DB archive status, and move archive/unarchive to a `ConversationArchiveStore` like pin/read.
- NPE while building notifications when a conversation is deleted mid-sync. Fix: skip the orphaned message.

inthewaves (approved):

PR #163, opened 2026-06-25
Forward Message Screen: Java/XML to Kotlin/Compose migration

m4pl:
Migrated the Forward Message flow from Java + XML to Kotlin + Compose, reusing the shared `ConversationPickerScreen` to cut duplicated code.

### Screenshots

| Before | After |
|--------|-------|
|Screenshot_20260625_022441|Screenshot_20260625_022959|


src/com/android/messaging/ui/conversationpicker/host/forward/ForwardMessageActivity.kt
```diff
@@ -0,0 +1,81 @@
...
+                    )
+                }
+
+                ConversationPickerScreen(
+                    effectHandler = effectHandler,
+                    onNavigateBack = ::finish,
+                    allowMultiSelect = true,
+                    isInitialDraftLoading = false,
+                    initialDraft = draft,
+                )
```
inthewaves, line 79: The ConversationPickerScreen hardcodes strings such as `share_intent_activity_label` and `share_recent_conversations_title`; those parts should probably be customized / use custom strings for this forwarding path

e.g. at the bottom of the picker, it says "Enter a contact name or phone number to start a new message" which doesn't have the context of forwarding message. The title could be customized to indicate that we are forwarding messages

inthewaves (changes requested):

inthewaves (approved):

PR #166, opened 2026-06-25
Rewrite low storage dialog to Compose

RankoR:


src/com/android/messaging/ui/smsstoragelow/SmsStorageLowWarningScreen.kt
```diff
@@ -0,0 +1,519 @@
...
+) {
+    val uiState by screenModel.uiState.collectAsStateWithLifecycle()
+
+    LaunchedEffect(screenModel, onFinish) {
+        screenModel.effects.collect { effect ->
+            when (effect) {
+                Effect.Finish -> onFinish()
+            }
+        }
+    }
```
inthewaves: We should use `rememberUpdatedState` for `onFinish` instead of using it as a `LaunchedEffect` key

inthewaves (changes requested):

src/com/android/messaging/ui/smsstoragelow/SmsStorageLowWarningScreen.kt
```diff
@@ -0,0 +1,519 @@
...
+) {
+    Surface(
+        modifier = Modifier
+            .fillMaxWidth()
+            .testTag(tag = smsStorageLowWarningActionTestTag(action = action))
+            .clickable(
+                enabled = enabled,
+                role = Role.Button,
+                onClick = onClick,
+            ),
```
m4pl: Ripple/selection isn't clipped to the rounded shape and bleeds past the corners

src/com/android/messaging/ui/smsstoragelow/SmsStorageLowWarningViewModel.kt
```diff
@@ -0,0 +1,177 @@
...
+                .catch { exception ->
+                    LogUtil.e(TAG, "Failed to release SMS storage", exception)
+                }
+                .onCompletion {
+                    _uiState.update { it.copy(isProcessing = false) }
+                    emitEffect(effect = Effect.Finish)
+                }
+                .collect {
+                    clearSelectedAction()
+                }
```
m4pl: Clearing the selection here flashes the choice dialog for a moment before the screen finishes

src/com/android/messaging/ui/smsstoragelow/SmsStorageLowWarningScreen.kt
```diff
@@ -0,0 +1,519 @@
...
+
+@Composable
+private fun SmsStorageLowWarningDialogSurface(
+    testTag: String,
+    modifier: Modifier = Modifier,
+    content: @Composable () -> Unit,
+) {
+    Surface(
+        modifier = modifier
+            .heightIn(max = DialogMaxHeight)
```
m4pl: Do we really need this height limit?

m4pl (commented):
Maybe we could add a way to open this dialog through the Debug settings?

inthewaves:
I'm actually not sure if this dialog with the clear data options is really needed or appropriate anymore. For example, Google Messages just tells users to clear up storage in settings to keep sending messages:

```
<string name="exhausted_storage_space_notification_and_dialog_title">To keep using Messages, free up space</string>
<string name="exhausted_storage_space_notification_body">On your device, go to Settings : Storage and free up space</string>
```

thestinger:
I agree with getting rid of this beyond a warning.

inthewaves (approved):

PR #176, opened 2026-07-02
Archived Conversation List Screen: Java/XML to Kotlin/Compose migration

m4pl:
Migrated the Archived Conversation List flow from Java + XML to Kotlin + Compose. Mostly moved files around for reuse and removed all legacy Java code related to the Conversation List. UI is the same as for the Conversation List Screen.


m4pl:
Waiting for the [Conversation List Screen PR](https://github.com/GrapheneOS/Messaging/pull/162) to be merged

AndroidManifest.xml
inthewaves: This is an exported activity, so if we change its name, we would need an `activity-alias` for the old name

Since this is actually the activity with intent filter for `android.intent.action.MAIN`, missing the alias would likely break existing shortcuts for the Messaging app in the launcher if the launcher stores exact resolved launch components

src/com/android/messaging/ui/conversationlist/archived/ArchivedConversationListScreen.kt
```diff
@@ -0,0 +1,341 @@
...
+    onNavigateBack: () -> Unit,
+    modifier: Modifier = Modifier,
+    screenModel: ArchivedConversationListScreenModel =
+        viewModel<ArchivedConversationListViewModel>(),
+) {
+    val uiState by screenModel.uiState.collectAsStateWithLifecycle()
+    val listState = rememberLazyListState()
+    val snackbarHostState = remember { SnackbarHostState() }
+
+    var pendingDelete by rememberSaveable { mutableStateOf(false) }
```
inthewaves: probably doesn't need to be saveable, especially when the ViewModel's `uiState` is not using anything saveable itself across process death (e.g. a `SavedStateHandle`)

src/com/android/messaging/ui/conversationlist/archived/ArchivedConversationListViewModel.kt
```diff
@@ -0,0 +1,261 @@
...
+        }
+
+        selectedItems.forEach { item ->
+            conversationsRepository.deleteConversation(
+                conversationId = item.conversationId,
+                cutoffTimestamp = item.latestMessage.timestamp,
+            )
+        }
+        selectionDelegate.clear()
+    }
```
inthewaves, line 223: Unarchive, undo, and delete are implemented directly in this archived ViewModel. The normal conversation list routes the same class of mutations through `ConversationListActionsDelegate` from its ViewModel. 

That gives conversation list mutations two owners. The delegate normalizes IDs, owns coroutine launch/scope behavior, and emits snackbar status effects, while this path launches directly from `viewModelScope`, calls the repository directly and emits archived-specific effects.

That split makes future policy changes easy to miss in one path, such as delete eligibility/default-SMS gating, failure handling, or snackbar/undo behavior. We should move the shared mutation policy into a common delegate/use case used by both inbox and archived, leaving only screen-specific UI effects in each ViewModel

src/com/android/messaging/ui/conversationlist/archived/mapper/ArchivedConversationListUiStateMapper.kt
```diff
@@ -0,0 +1,56 @@
...
+            items.isNotEmpty() -> {
+                ConversationListContentUiState.Items(
+                    items = items,
+                    restoredConversationIds = snapshot.restoredConversationIds,
+                )
+            }
+
+            else -> {
+                ConversationListContentUiState.Empty
+            }
```
inthewaves: This duplicates the inbox snapshot-to-content mapping from `src/com/android/messaging/ui/conversationlist/chats/mapper/ConversationListUiStateMapper.kt` but omits the `!snapshot.hasFirstSyncCompleted -> WaitingForSync` branch

As both screens share `ConversationListSnapshot` and `ConversationListContentUiState`, this common part should probably have one owner. Maybe also makes sense to have archived conversations show waiting for sync message too

src/com/android/messaging/ui/conversationlist/chats/ConversationListContent.kt
```diff
@@ -0,0 +1,160 @@
...
+                    listState = listState,
+                    isSelectionMode = isSelectionMode,
+                    scaffoldContentPadding = scaffoldContentPadding,
+                    fabBottomReserve = fabBottomReserve,
+                    pinAnimationController = pinAnimationController,
+                    callbacks = inboxCallbacks(onAction),
+                    swipeActions = { item -> inboxSwipeActions(item.conversationId, onAction) },
+                )
+            }
+        }
```
inthewaves, line 86: See other review comment about `inboxCallbacks`, `inboxSwipeActions`, etc. 

The move from a single lambda to a data class holding lambdas is currently contributing to all conversation row items recomposing (from Android Studio Layout Inspector) even when only 1 conversation changes (although https://github.com/GrapheneOS/Messaging/issues/190 also causes all rows to recompose in this case)

src/com/android/messaging/ui/conversationlist/common/list/ConversationListItems.kt
```diff
@@ -73,70 +51,31 @@ private val ListVerticalSpacing = 2.dp
...
-                    pinAnimationController = pinAnimationController,
-                )
-            }
-        }
-    }
-}
+internal data class ConversationListSwipeActions(
+    val startToEnd: ConversationSwipeAction?,
+    val endToStart: ConversationSwipeAction?,
+)
```
inthewaves: Putting lambdas into data classes changes Compose compiler behavior

Before this refactor, the shared list mostly passed the single `onAction` event sink through and built row-specific lambda expressions directly in composable code. That shape lets the Compose compiler apply lambda memoization under strong skipping (https://github.com/JetBrains/kotlin/blob/master/plugins/compose/design/strong-skipping.md#lambda-memoization)

After the refactor, those lambdas are wrapped in `ConversationListItemCallbacks`, a data class returned from helper functions like `inboxCallbacks(onAction)`. The compiler does not automatically memoize that data-class holder. Its generated `equals()` compares the lambda fields, and newly allocated lambda instances normally compare unequal, so the holder is likely seen as changed on each recomposition.

We could restructure the shared list API so every hot-path parameter is either structural-equality data or a single function type:

```kotlin
// common/list
@Immutable
internal sealed interface ConversationListItemEvent {
    data class Clicked(val conversationId: String) : ConversationListItemEvent
    data class LongClicked(val conversationId: String) : ConversationListItemEvent
    data class AvatarMessageClicked(val conversationId: String) : ConversationListItemEvent
    data class AvatarCallClicked(val destination: String) : ConversationListItemEvent
    data class AvatarContactClicked(val item: ConversationListItemUiModel) : ConversationListItemEvent
    data class AvatarInfoClicked(val conversationId: String) : ConversationListItemEvent
    data class Swiped(
        val conversationId: String,
        val kind: ConversationSwipeKind,
    ) : ConversationListItemEvent
}

@Immutable
internal data class ConversationListSwipeSpec(
    val startToEnd: ConversationSwipeKind?, // kind carries background + semantics
    val endToStart: ConversationSwipeKind?,
)

```

Each screen would have a top-level constant and one mapping function

```kotlin
private val InboxSwipeSpec = ConversationListSwipeSpec(
    startToEnd = ConversationSwipeKind.ToggleRead,
    endToStart = ConversationSwipeKind.Archive,
)

ConversationListItems(
    items = content.items,
    swipeSpec = InboxSwipeSpec,
    onItemEvent = { onAction(it.toInboxAction()) }, // compiler-memoized lambda
)

private fun ConversationListItemEvent.toInboxAction(): Action = when (this) { /* ... */ }
```

inthewaves (changes requested):

inthewaves (approved):

thestinger:
This is blocked due to a conflict according to GitHub.

inthewaves (approved):

PR #184, opened 2026-07-08
VCard Detail Screen: Java/XML to Kotlin/Compose migration

m4pl:
Migrated the VCard Detail flow and inline vCard cards from Java + XML to Kotlin + Compose, and removed the legacy vCard Java code.

Also swapped the old `MediaResourceManager`/`VCARD_CACHE` caching for a new `VCardEntryRepository` on top of a Kotlin `VCardParser`. The old cache could show stale data for vCards backed by a contact, so those now refresh via a `ContentObserver` when the contact is edited.

### Screenshots

| Before | After |
|--------|-------|
|Screenshot_20260708_032321|Screenshot_20260708_031733|

### Video
https://github.com/user-attachments/assets/e7489420-9053-4bd8-88c4-9d160bdac07a


src/com/android/messaging/data/vcard/repository/VCardEntryRepository.kt
```diff
@@ -0,0 +1,211 @@
...
+                shouldCache = shouldCache,
+            )
+            entries
+        }
+
+        result.onSuccess { entries ->
+            parseHandle.deferred.complete(entries)
+        }.onFailure { throwable ->
+            parseHandle.deferred.completeExceptionally(throwable)
+        }
```
inthewaves: Need to be careful with using broad catching functions when the body is a suspend lambda/block calling suspendable (and thus possibly cancellable) functions. Kotlin coroutine cancellation is done through `CancellationExceptions` which should never be caught

src/com/android/messaging/ui/conversationsettings/screen/mapper/ConversationSettingsUiStateMapper.kt
```diff
@@ -76,19 +74,15 @@ internal class ConversationSettingsUiStateMapperImpl @Inject constructor(
...
+            hasFullName -> fullName
+            else -> participant.sendDestination.orEmpty()
         }
         val details = when {
-            fullName.isNullOrEmpty() || participant.isUnknownSender -> null
-            else -> participant.sendDestination?.let {
-                bidiFormatter.unicodeWrap(it, LTR)
-            }
+            hasFullName && !participant.isUnknownSender -> participant.sendDestination
+            else -> null
```
inthewaves: We need to retain the `fullName.isNullOrEmpty()` instead of `fullName.isEmpty()`. This currently results in an NPE when opening conversations settings for an unsaved number which would have a `null` `fullName` (should add a test for this too)

src/com/android/messaging/data/vcarddetail/repository/VCardDetailRepository.kt
```diff
@@ -0,0 +1,55 @@
...
+            emit(VCardDetailResult.Loading)
+
+            vCardEntryRepository
+                .observeEntries(
+                    vCardUri = vCardUri,
+                    refreshes = refreshes,
+                )
+                .map(::mapResult)
+                .collect(::emit)
+        }
```
inthewaves: should probably specify dispatcher via flowOn, otherwise this could flow and thus do parsing/bitmap decoding on main thread

src/com/android/messaging/ui/appsettings/subscription/ui/SubscriptionSettingsScreen.kt
```diff
@@ -163,7 +164,7 @@ private fun LazyListScope.mmsSettingsItems(
     item(key = "phone_number") {
         SettingsClickableItem(
             title = stringResource(R.string.mms_phone_number_pref_title),
-            summary = subscriptionSettings.displayDetail,
+            summary = subscriptionSettings.displayDetail.asLtrText(),
```
inthewaves, line 167: this should probably also be applied to SettingsMainScreen.kt:65 which also has `subscription.displayDetail`

src/com/android/messaging/data/vcard/repository/VCardEntryRepository.kt
```diff
@@ -0,0 +1,211 @@
...
+        return !vCardUri.toUri().isContactVCardUri()
+    }
+
+    private fun Uri.isContactVCardUri(): Boolean {
+        val contactVCardUri = Contacts.CONTENT_VCARD_URI
+
+        return scheme == contactVCardUri.scheme &&
+            authority == contactVCardUri.authority &&
+            pathSegments.take(contactVCardUri.pathSegments.size) == contactVCardUri.pathSegments
+    }
```
inthewaves, line 141: We should probably also not cache `Contacts.CONTENT_MULTI_VCARD_URI`

src/com/android/messaging/data/vcard/mapper/VCardEntrySummarizer.kt
```diff
@@ -0,0 +1,98 @@
...
+) : VCardEntrySummarizer {
+
+    override fun displayName(entry: CustomVCardEntry): String? {
+        val name = entry.displayName ?: run {
+            entry.consolidateFields()
+            entry.displayName
+        }
+
+        return name?.takeIf(String::isNotBlank)
+    }
```
inthewaves, line 25: This seems to be doing hidden mutations in `entry.consolidateFields()` which isn't good practice especially if this can be called concurrently; it's probably better to do it once in e.g. VCardParser

src/com/android/messaging/ui/vcarddetail/common/VCardContactCard.kt
```diff
@@ -0,0 +1,208 @@
...
+            .fillMaxWidth()
+            .combinedClickable(
+                onClick = onClick,
+                onLongClick = onLongClick,
+            )
+            .padding(
+                horizontal = RowHorizontalPadding,
+                vertical = RowVerticalPadding,
+            ),
+        verticalArrangement = Arrangement.spacedBy(2.dp),
```
inthewaves, line 145: Note that the `VCardDetailViewModel` can drop clicks for `VCardFieldAction.None`, but semantically it will still show a click action for e.g. TalkBack

src/com/android/messaging/ui/vcarddetail/common/VCardContactCard.kt
```diff
@@ -0,0 +1,208 @@
...
+    displayName: String?,
+    normalizedDestination: String?,
+    avatarPhoto: VCardAvatarPhoto?,
+) {
+    val displayText = displayName?.asLtrText().orEmpty()
+    val avatarImage = remember(avatarPhoto) {
+        avatarPhoto
+            ?.asReadOnlyByteBuffer()
+            ?.let(ParticipantAvatarImage::Bytes)
+    }
```
inthewaves: Repeated code and could be refactored

src/com/android/messaging/ui/conversation/attachment/ui/ConversationVCardAttachmentCardContent.kt
```diff
@@ -30,22 +32,30 @@ private const val PREVIEW_LONG_LOCATION_SUBTITLE =
...
     titleTextResId: Int?,
+    normalizedDestination: String?,
     subtitleText: String?,
     subtitleTextResId: Int?,
 ) {
+    val avatarImage = remember(avatarPhoto) {
+        avatarPhoto
+            ?.asReadOnlyByteBuffer()
+            ?.let(ParticipantAvatarImage::Bytes)
+    }
```
inthewaves: Repeated code and could be refactored

src/com/android/messaging/ui/conversation/composer/ui/ConversationAttachmentPreview.kt
```diff
@@ -113,16 +115,23 @@ private fun ResolvedAttachmentPreviewItem(
 
     when (attachment) {
         is ComposerAttachmentUiModel.Resolved.VCard -> {
+            val avatarImage = remember(attachment.vCardUiModel.avatarPhoto) {
+                attachment.vCardUiModel.avatarPhoto
+                    ?.asReadOnlyByteBuffer()
+                    ?.let(ParticipantAvatarImage::Bytes)
+            }
```
inthewaves: Repeated code and could be refactored

inthewaves (changes requested):

src/com/android/messaging/ui/conversation/messages/delegate/ConversationMessagesDelegate.kt
```diff
@@ -59,8 +64,13 @@ internal class ConversationMessagesDelegateImpl @Inject constructor(
...
 
         scope.launch(defaultDispatcher) {
+            var currentConversationId: String? = null
+
             conversationIdFlow.collectLatest { conversationId ->
-                _state.value = ConversationMessagesUiState.Loading
+                if (conversationId != currentConversationId) {
+                    currentConversationId = conversationId
+                    _state.value = ConversationMessagesUiState.Loading
+                }
```
inthewaves: I don't think `currentConversationId` and the equality check is needed.  `conversationIdFlow` is a `StateFlow`, which never emits consecutive equal values

`_state` is also a `StateFlow`, so setting `Loading` unconditionally is also a no-op when the state is already Loading, for the same reason

src/com/android/messaging/ui/vcarddetail/screen/VCardDetailScreen.kt
```diff
@@ -0,0 +1,197 @@
...
+            top = ScreenContentPadding,
+            bottom = ScreenContentPadding + scaffoldContentPadding.calculateBottomPadding(),
+            start = ScreenContentPadding + horizontalInsets.calculateStartPadding(layoutDirection),
+            end = ScreenContentPadding + horizontalInsets.calculateEndPadding(layoutDirection),
+        ),
+    ) {
+        itemsIndexed(
+            items = contacts,
+            key = { index, _ -> index },
+        ) { _, contact ->
```
inthewaves: This `key = { index, _ -> index }` is redundant since `itemsIndexed` documentation already says:

> If null is passed the position in the list will represent the key.

src/com/android/messaging/ui/vcarddetail/screen/mapper/VCardDetailUiStateMapper.kt
```diff
@@ -0,0 +1,66 @@
...
+            fields = contact.fields
+                .map(::mapField)
+                .toImmutableList(),
+        )
+    }
+
+    private fun mapField(field: VCardField): VCardFieldUiModel {
+        return VCardFieldUiModel(
+            value = field.value,
+            displayValue = field.value,
```
inthewaves: It seems like `value` is always equal to `displayValue`

src/com/android/messaging/data/vcard/repository/VCardEntryRepository.kt
```diff
@@ -0,0 +1,211 @@
...
+    private fun contactChangeEvents(): Flow<Unit> {
+        return callbackFlow {
+            val observer = object : ContentObserver(null) {
+                override fun onChange(selfChange: Boolean) {
+                    trySend(Unit)
+                }
+            }
+
+            contentResolver.registerContentObserver(Contacts.CONTENT_URI, true, observer)
+            trySend(Unit)
```
inthewaves, line 120: Note that this `trySend(Unit)` will trigger a full parse after first observer registration and can cause double parsing on open; see https://github.com/GrapheneOS/Messaging/pull/184#discussion_r3550344550

src/com/android/messaging/ui/conversation/screen/ConversationViewModel.kt
```diff
@@ -788,6 +788,8 @@ internal class ConversationViewModel @Inject constructor(
     }
 
     override fun onScreenForegrounded(cancelNotification: Boolean) {
+        conversationComposerAttachmentsDelegate.refresh()
+        conversationMessagesDelegate.refresh()
```
inthewaves, line 792: Note that this will trigger a vCards parse . This will do uncached contact vCards parsing twice on first open, since the repository's initial observer registration does its own `trySend(Unit)` from the initial `refreshEvents`

e.g. put a log inside of `VCardEntryRepositoryImpl.parseShared` before the `val result = runCatching` call, and 2 log lines will be printed

inthewaves (changes requested):

src/com/android/messaging/data/vcard/mapper/VCardEntrySummarizer.kt
inthewaves: Review lines got messed up, but it's for this:

```kotlin
override fun displayName(entry: CustomVCardEntry): String? {
        val name = entry.displayName ?: run {
            entry.consolidateFields() // <--
            entry.displayName
        }

        return name?.takeIf(String::isNotBlank)
    }
```

src/com/android/messaging/data/vcard/repository/VCardEntryRepository.kt
m4pl: Dropped the in-flight dedup, so there's no broad catch anymore. `getEntries` just parses directly and relies on the cache.

src/com/android/messaging/data/vcard/mapper/VCardEntrySummarizer.kt
m4pl: Already handled by `CustomVCardEntryConstructor.onEntryEnded()` before `onEntryCreated()`, so I removed it here. displayName is now just a read.

inthewaves:
LGTM, just needs to fix conflicts with main branch

inthewaves (approved):

PR #192, opened 2026-07-09
Modernise GifTranscoder JNI files

mikomikotaishi:
This PR addresses the following in the native JNI parts of `GifTranscoder`:

- Set an explicit C++ version (C++17), which is the NDK default
- Replace macros with `constexpr` functions and constants where possible
- Add `[[nodiscard]]` attributes, to explicitly highlight when a function call's return goes unused
- Add `noexcept` specifiers
- `computeAverage()` stored channels in plain `char`; any channel above `0x7f` would become negative and upon left-shifting, causes undefined behaviour; instead, cast to `unsigned char` (or `uint8_t`).
- `makeColorARGB()` takes `uint8_t` rather than `int` so an out-of-range channel cannot be silently passed.
- Replace C-style casts `(T)` with `static_cast<T>` and `reinterpret_cast<T>`
- Replace `NULL` with `nullptr`
- Replace a `typedef` with a `using` alias
- Replace bodyless constructor/destructor with `= default`
- Replace `int[4]` with `std::array<int, 4>`
- Replace the C `::timespec` and `::clock_gettime()` with `std::chrono::steady_clock`
- Explicitly disable copy constructor and copy assignment operators on `GifFilesCloser`
- Add a regression test, `com.android.messaging.util.GifTranscoderTest`, a 4x2 gif; the grey block is the assertion that catches the bug, while the white block is the control.

Note that the first commit is a refactor; the second commit is the actual behavioural change.


mikomikotaishi:
@m4pl @RankoR @thestinger I'd like to request a PR review (I do not have access to assign reviewers)

RankoR (changes requested):
Thanks for your contribution.

It's good in general, but could you please do these adjustments?

1. Update the PR description and commit message to explicitly mention the x86_64 signed-char undefined behavior and color-corruption fix
2. Add a regression test using color channels above 0x7f. An end-to-end test that transcodes the small 4×2 fixture and checks the resulting two pixels would catch the old implementation
3. Keep the functional fix in a separate commit from the mechanical modernization

mikomikotaishi:
Where would you like the test to go?

RankoR:
@mikomikotaishi you can add something like `app/src/androidTest/java/com/android/messaging/util/GifTranscoderTest.kt`, put a GIF in `app/src/androidTest/assets/gif_transcoder/high_color_channels_4x2.gif` and load it from the test.

mikomikotaishi:
Hi, sorry about the wait. Let me add some tests as requested

mikomikotaishi:
@RankoR Could you look at the state of the PR again?

RankoR (approved):

jni/GifTranscoder.cpp
```diff
@@ -59,19 +43,64 @@ static const int64_t kGifMaxArea = 4LL * 1024 * 1024;
...
+
+[[nodiscard]]
+constexpr ColorARGB makeColorARGB(uint8_t a, uint8_t r, uint8_t g, uint8_t b) noexcept {
+    return (static_cast<ColorARGB>(a) << 24) | (static_cast<ColorARGB>(r) << 16) |
+           (static_cast<ColorARGB>(g) << 8) | static_cast<ColorARGB>(b);
+}
+
+constexpr uint32_t MAX_COLOR_DISTANCE = 255 * 255 * 255;
+
+constexpr size_t GIF_MAX_AREA = 4LL * 1024 * 1024;
```
inthewaves: nit: since we're more or less following Google's style guide, their style guide says to use names like `kMaxColorDistance` for constexpr and const (https://google.github.io/styleguide/cppguide.html#Constant_Names). `ALL_CAPS` is for macros (https://google.github.io/styleguide/cppguide.html#Macro_Names)

jni/GifTranscoder.cpp
```diff
@@ -59,19 +43,64 @@ static const int64_t kGifMaxArea = 4LL * 1024 * 1024;
...
-    struct timespec res;
-    clock_gettime(CLOCK_REALTIME, &res);
-    return 1000.0 * res.tv_sec + (double) res.tv_nsec / 1e6;
+[[nodiscard]]
+constexpr int square(int a) noexcept {
+    return a * a;
+}
+
+// GIF does not support partial transparency, so our alpha channels are always 0x0 or 0xff.
+constexpr ColorARGB TRANSPARENT = 0x0;
```
inthewaves: https://google.github.io/styleguide/cppguide.html#Constant_Names

jni/GifTranscoder.cpp
```diff
@@ -560,16 +586,16 @@ jboolean transcode(JNIEnv* env, jobject clazz, jstring filePath, jstring outFile
     return (gifCode == GIF_OK);
 }
 
-const char *kClassPathName = "com/android/messaging/util/GifTranscoder";
+constexpr char CLASS_PATH_NAME[] = "com/android/messaging/util/GifTranscoder";
```
inthewaves: https://google.github.io/styleguide/cppguide.html#Constant_Names

jni/GifTranscoder.cpp
```diff
@@ -560,16 +586,16 @@ jboolean transcode(JNIEnv* env, jobject clazz, jstring filePath, jstring outFile
...
 
-JNINativeMethod kMethods[] = {
-        { "transcodeInternal", "(Ljava/lang/String;Ljava/lang/String;)Z", (void*)transcode },
+const JNINativeMethod METHODS[] = {
+    { "transcodeInternal", "(Ljava/lang/String;Ljava/lang/String;)Z", reinterpret_cast<void*>(transcode) },
 };
```
inthewaves: https://google.github.io/styleguide/cppguide.html#Constant_Names

jni/GifTranscoder.cpp
```diff
@@ -14,33 +14,17 @@
...
-#define RED(color)   (((color) >> 16) & 0xff)
-#define GREEN(color) (((color) >>  8) & 0xff)
-#define BLUE(color)  (((color) >>  0) & 0xff)
-
-#define MAKE_COLOR_ARGB(a, r, g, b) \
-    ((a) << 24 | (r) << 16 | (g) << 8 | (b))
-
-#define MAX_COLOR_DISTANCE (255 * 255 * 255)
-
-static const int64_t kGifMaxArea = 4LL * 1024 * 1024;
```
inthewaves, line 42: Note that this was already following https://google.github.io/styleguide/cppguide.html#Constant_Names, but it seems Google was inconsistent before

inthewaves (changes requested):

mikomikotaishi:
That's fine. I can revert the `ALL_CAPS_CONSTANTS` to `kConstants` style. (I just copied `ALL_CAPS_CONSTANTS` because this was a Java/Kotlin codebase and it seemed logical to apply the same convention here.)

inthewaves (approved):

PR #193, opened 2026-07-09
Fix crash receiving SMS when the local database insert fails

thomasbuilds:
Commit 641e11b dropped the null guard around the received-message log statement while keeping the catch block that swallows database failures; commit 91c7eb7d had introduced the two together. If anything throws inside the transaction before the MessageData is constructed - such as a transient SQLiteException from the participant lookup - message stays null and the unconditional message.getMessageId() call throws an uncaught NullPointerException that kills the process, skipping the notification and pending-message processing even though the message was already written to telephony.

Restore the guard so these failures degrade gracefully again.


RankoR (changes requested):
LGTM, but please add a test.

RankoR (approved):

PR #194, opened 2026-07-09
Fix crash when declining a call with a quick response

thomasbuilds:
Commit 9aec6f6b replaced the argument-less notification refresh with BugleNotifications.updateWithInlineReply(conversationId, message), which is also reached from the recipients-only RESPOND_VIA_MESSAGE path where conversationId is null. Telecom invokes this exported service whenever the user rejects an incoming call with a quick-response text, and the null conversationId crashes on the non-null channelId parameter of NotificationChannelUtil.getActiveNotification. Since the service sets intent redelivery, the crash also redelivers the intent and re-sends the reply, causing duplicate texts alongside the crash loop.

Move the call into the branch where the conversation is known. The notification inline-reply path always supplies a conversation id, so its behavior is unchanged, and the recipients-only path has no active conversation notification to update in the first place.

### DIFF
--- a/app/src/test/kotlin/com/android/messaging/datamodel/NoConfirmationSmsSendServiceRespondViaMessageTest.kt
+++ b/app/src/test/kotlin/com/android/messaging/datamodel/NoConfirmationSmsSendServiceRespondViaMessageTest.kt
@@ -1,0 +1,101 @@
+package com.android.messaging.datamodel
+
+import android.content.Intent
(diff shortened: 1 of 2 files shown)


RankoR (changes requested):
LGTM, But please add a regression test (we already have Robolectric in the project - you can use it).

RankoR (approved):

PR #195, opened 2026-07-10
Debug MMS Config Screen: Java/XML to Kotlin/Compose migration

m4pl:
Screenshot_20260710_185056


src/com/android/messaging/data/debugmmsconfig/repository/MmsConfigRepository.kt
```diff
@@ -0,0 +1,97 @@
...
+        subId: SubId,
+        key: String,
+        keyType: MmsConfigKeyType,
+        value: String,
+    ) {
+        withContext(ioDispatcher) {
+            MmsConfig.get(subId.value)
+                .update(keyType.rawType, key, value)
+        }
+    }
```
inthewaves, line 55: This is in a suspend function, so concurrent execution is possible if callers are calling this on multiple threads/coroutines. 

`ioDispatcher` has many threads but `MmsConfig` / `BugleCarrierConfigValuesLoader` is generally not safe from concurrent mutations. `MmsConfig#update` calls `BugleCarrierConfigValuesLoader.update(mValues, type, key, value)` modifies a `Bundle` which is not safe from concurrent mutations

Also the `MmsConfigRepository` uses `@Reusable` so multiple instances can exist and those can also concurrently modify the global per-subId `MmsConfig` instance stored in `MmsConfig’`s static cache.

src/com/android/messaging/di/debugmmsconfig/DebugMmsConfigBindsModule.kt
```diff
@@ -0,0 +1,28 @@
...
+
+@Module
+@InstallIn(SingletonComponent::class)
+internal abstract class DebugMmsConfigBindsModule {
+
+    @Binds
+    @Reusable
+    abstract fun bindMmsConfigRepository(
+        impl: MmsConfigRepositoryImpl,
+    ): MmsConfigRepository
```
inthewaves, line 21: The repository is bound with `@Reusable`, so Dagger does not guarantee a single instance. Multiple   instances can theoretically exist, although current usage appears to inject `MmsConfigRepositoryImpl` only through `DebugMmsConfigViewModel`.

https://dagger.dev/dev-guide/

> There is no guarantee that the component will call the binding only once, so applying `@Reusable` to bindings that return mutable objects, or objects where it’s important to refer to the same instance, is dangerous. It’s safe to use `@Reusable` for immutable objects that you would leave unscoped if you didn’t care how many times they were allocated.

`MmsConfigRepositoryImpl` mutates the global per-subId `MmsConfig` instance stored in `MmsConfig’`s static cache

inthewaves (changes requested):

src/com/android/messaging/di/debugmmsconfig/DebugMmsConfigBindsModule.kt
m4pl: Replaced `IoDispatcher` with `MessagingDbDispatcher`, which solves both problems. The only downside is the dispatcher's name. I don't think it's worth creating a new one for this debug screen. What do you think, @inthewaves?

inthewaves: Yeah if it's just for debug it's not worth the hassle I think

inthewaves (approved):

PR #197, opened 2026-07-11
Show MMS subject in tap to download message

m4pl:
image


src/com/android/messaging/ui/conversation/messages/ui/message/ConversationMessageBubble.kt
```diff
@@ -447,6 +418,22 @@ private fun ConversationMessageBody(
...
 }
 
+@Composable
+private fun ConversationMessageSubject(
+    subjectText: String?,
+    modifier: Modifier = Modifier,
+) {
+    if (subjectText == null) {
+        return
+    }
```
inthewaves, line 428: `subjectText` could also be blank too; https://github.com/GrapheneOS/Messaging/blob/main/src/com/android/messaging/sms/MmsSubjectSanitizer.kt doesn't clean blanks (it probably should)

inthewaves (changes requested):

inthewaves (approved):

PR #200, opened 2026-07-12
Fix secondary user SMS notification hiding sender and message

### LINKED ISSUE #75: SMS notifications require tap
### Description  
When you receive a SMS as a secondary user you have to tap the notification in order to read it. This is poor usability and should be improved.

### Steps to Reproduce
1. Create a secondary user
2. Log in as the secondary user, keep the screen unlocked
3. receive a SMS. There will be a notification but it will hide its content and the sender unless you tap on it.

### Expected Behaviour
The behaviour should be the same as when you receive the SMS while being logged in as the owner. As an owner the notification directly shows the sender and the message.

### Actual Behaviour
screenshot of notification

### Environment
- OS: GrapheneOS build 2025073000, stable release as of today
- SMS app bundled with GrapheneOS, version 11

m4pl:
Closes https://github.com/GrapheneOS/Messaging/issues/75

Fixed: secondary users saw a generic "New message" instead of the sender and text. The notification now shows them directly.

Also: receivers moved to Kotlin, receive/parse/notify logic split into DI collaborators, phone number formatting consolidated.

How to check on device:
1. As owner, enable "Phone calls & SMS" for a secondary user.
2. Switch to that user, open the app.
3. Text your own number. The notification shows the sender and message.


src/com/android/messaging/receiver/SmsReceiver.kt
```diff
@@ -0,0 +1,71 @@
...
+class SmsReceiver : BroadcastReceiver() {
+
+    override fun onReceive(context: Context, intent: Intent) {
+        LogUtil.v(LogUtil.BUGLE_TAG, "SmsReceiver.onReceive $intent")
+
+        if (!PhoneUtils.getDefault().isSmsEnabled || !OsUtil.isSecondaryUser()) {
+            return
+        }
+
+        if (intent.action == Sms.Intents.SMS_RECEIVED_ACTION) {
```
RankoR, line 25: `android.provider.Telephony.MMS_DOWNLOADED` is not handled there, so user will not receive a notification

src/com/android/messaging/receiver/SmsReceiver.kt
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    private fun handleSecondaryUserSmsReceived(
+        context: Context,
+        intent: Intent,
+        entryPoint: IncomingSmsEntryPoint,
+    ) {
+        // The contact lookup performs a synchronous query, so resolve off the main thread.
+        val pendingResult = goAsync()
+        entryPoint.applicationScope().launch(entryPoint.ioDispatcher()) {
```
RankoR, line 37: Concurrent broadcasts can display an older SMS as the latest notification.
Every broadcast starts an independent `Dispatchers.IO` coroutine, so:

1. SMS A starts a slow contacts lookup
2. SMS B arrives later and resolves quickly
3. B posts its notification
4. A completes and overwrites B using the shared notification ID

src/com/android/messaging/receiver/SmsReceiver.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    override fun onReceive(context: Context, intent: Intent) {
+        LogUtil.v(LogUtil.BUGLE_TAG, "SmsReceiver.onReceive $intent")
+
+        if (!PhoneUtils.getDefault().isSmsEnabled || !OsUtil.isSecondaryUser()) {
+            return
+        }
+
+        if (intent.action == Sms.Intents.SMS_RECEIVED_ACTION) {
+            handleSecondaryUserSmsReceived(context, intent, entryPoint(context))
```
RankoR, line 26: Would it be more reliable to handle it in a `Service` instead of short-living Broadcast, given that we're querying db with unpredictable delays?

src/com/android/messaging/receiver/SmsReceiver.kt
```diff
@@ -0,0 +1,71 @@
...
+            SendStatusReceiver.NO_ERROR_CODE,
+        )
+
+        return entryPoint.secondaryUserMessageResolver().resolve(
+            address = values.getAsString(Sms.ADDRESS),
+            body = values.getAsString(Sms.BODY),
+        )
+    }
+
+    private fun entryPoint(context: Context): IncomingSmsEntryPoint {
```
RankoR, line 65: Why not regular injection?

app/src/test/kotlin/com/android/messaging/data/secondaryuser/SecondaryUserMessageResolverTest.kt
```diff
@@ -0,0 +1,72 @@
...
+
+import com.android.messaging.data.phone.formatter.PhoneNumberFormatter
+import com.android.messaging.data.secondaryuser.model.SecondaryUserMessageInfo
+import io.mockk.every
+import io.mockk.mockk
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNull
+import org.junit.Test
+
+internal class SecondaryUserMessageResolverTest {
```
RankoR, line 12: Test coverage should be improved, currently tests are skipping lots of cases

RankoR (changes requested):

src/com/android/messaging/receiver/SmsReceiver.kt
m4pl: I checked and couldn't find this action broadcast anywhere. Messaging sends its own `SendStatusReceiver.MMS_DOWNLOADED` for MMS downloads, and incoming MMS comes through `WAP_PUSH_DELIVER`, so this branch never fired. @inthewaves, what do you think?

src/com/android/messaging/receiver/SmsReceiver.kt
m4pl: As far as I know the system gives a receiver ~10s, and a single contact lookup should finish well within that. Service/WorkManager feels like overkill here, and this worked fine before.

src/com/android/messaging/receiver/SmsReceiver.kt
m4pl: The receiver is created by the system, not by Hilt, so constructor injection isn't possible. The usual alternative is `@AndroidEntryPoint` field injection, but that requires calling `super.onReceive()`, and it doesn't compile because `BroadcastReceiver.onReceive` is abstract. So the remaining option is `EntryPointAccessors` to reach the Hilt graph.

RankoR: > AndroidEntryPoint

Yes, that's what I mean, not injecting the receiver itself, of course :)

> but that requires calling super.onReceive()

Ah, okay

RankoR (approved):

PR #203, opened 2026-07-13
Fix blocked conversation playing notification

### LINKED ISSUE #82: Blocking contacts does not block notification sounds
Profiles used: Owner, Secondary

Two GrapheneOS phones were used, both Pixel 7. Build 2025092700

## Steps:
1. Open Messaging app
2. Select a conversation
3. Tap three-dot menu and select People & Options
4. Block <contact name / number>
5. Send message **from** the contact that was blocked.

## Expected behavior:
Continue to receive messages *but without notifications* (including sound), and the conversation archived.

## Actual behavior:
Conversation is archived and displayed notifications were not received, as expected, but still received a notification sound.

## Additional Notes  

I have used Owner and secondary profiles with no effect. I tried removing the number from my contacts or changing the number format in case that mattered. I tried on both phones to block their respective numbers.

Deleting app Storage (Messaging > Storage & cache > Clear storage) has no effect on this bug.

I experimented with different Notifications settings. Disabling them, for all messages as well as individual contacts, silent notifications, removing it as a conversation (Messaging > conversation > People & Options > Notifications > Not a conversation). The only thing that disabled notification **sounds** was either turning off all notifications for Messages, or putting Notification volume at 0.

m4pl:
Closes https://github.com/GrapheneOS/Messaging/issues/82

This only reproduces while the blocked conversation itself is open on screen. On the list or in the background the observable soft-sound path no longer fires.

### DIFF
--- a/app/src/test/kotlin/com/android/messaging/datamodel/BugleNotificationsBlockedConversationTest.kt
+++ b/app/src/test/kotlin/com/android/messaging/datamodel/BugleNotificationsBlockedConversationTest.kt
@@ -1,0 +1,172 @@
+package com.android.messaging.datamodel
+
+import android.media.AudioManager
+import android.net.Uri
+import com.android.messaging.FactoryTestAccess
+import com.android.messaging.datamodel.data.ConversationListItemData
+import com.android.messaging.testutil.installTestFactory
+import com.android.messaging.util.RingtoneUtil
+import io.mockk.every
+import io.mockk.just
+import io.mockk.mockk
+import io.mockk.mockkStatic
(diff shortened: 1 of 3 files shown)


RankoR (changes requested):
It seems to be executed only if `MessageNotificationState.getNotificationState()` returns `null` (`BugleNotifications.java:157`). Non-null will bypass it.

It would be also nice to have tests for it.

sdsantos (approved):
I have been able to reproduce the issue and the changes fixed it.

Code looks good to me.

RankoR (approved):

PR #204, opened 2026-07-14
Direct secondary users to the Owner user for MMS

### LINKED ISSUE #170: MMS "tap to download" and auto-retrieve settings is shown in secondary users despite them being nonfunctional
The Messaging app currently doesn't download MMS in secondary users, e.g. see the various comments in  https://github.com/GrapheneOS/Messaging/blob/12/src/com/android/messaging/datamodel/data/MessageData.java#L593-L602

```java
    public boolean canDownloadMessage() {
        if (OsUtil.isSecondaryUser()) {
            // Secondary users can't download mms's. Mms's are downloaded by bugle running as the
            // primary user.
            return false;
        }
        // Can download if status is retrying auto/manual downloading
        return (mStatus == BUGLE_STATUS_INCOMING_RETRYING_MANUAL_DOWNLOAD ||
                mStatus == BUGLE_STATUS_INCOMING_RETRYING_AUTO_DOWNLOAD);
    }
```

For the new rewrite, undownloaded MMS messages will say "Tap to download" which do nothing when in a secondary user

Also the auto-retrieve MMS setting is a no-op because secondary user Messaging app doesn't download MMS. 

It should probably just refer users to the main/owner user for these operations

m4pl:
Closes https://github.com/GrapheneOS/Messaging/issues/170

Secondary users can't download MMS, so undownloaded messages now say "Download it from the Owner user" and the auto-retrieve switches are disabled with an explanation.

Seeder: dropped a fake `sms_message_uri`. Because it was set, sync treated the seeded rows as real telephony messages and deleted them, so they disappeared on device. `DownloadMmsAction`/`clearMmsStatus`: added a null-URI guard so a local-only row without a telephony URI doesn't try to download. Real messages always have a URI, so nothing changes for them.

### DIFF
--- a/app/src/test/kotlin/com/android/messaging/ui/conversation/messages/mapper/conversationmessage/ConversationMessageUiModelMapperMmsDownloadTest.kt
+++ b/app/src/test/kotlin/com/android/messaging/ui/conversation/messages/mapper/conversationmessage/ConversationMessageUiModelMapperMmsDownloadTest.kt
@@ -2,6 +2,8 @@ package com.android.messaging.ui.conversation.messages.mapper.conversationmessag
 
 import com.android.messaging.datamodel.data.MessageData
 import com.android.messaging.ui.conversation.messages.model.message.MmsDownloadUiModel
+import com.android.messaging.util.OsUtil
+import io.mockk.every
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNull
 import org.junit.Test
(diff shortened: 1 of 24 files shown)


app/src/test/kotlin/com/android/messaging/ui/conversation/messages/mapper/conversationmessage/ConversationMessageUiModelMapperMmsDownloadTest.kt
```diff
@@ -27,6 +27,7 @@ internal class ConversationMessageUiModelMapperMmsDownloadTest :
                 state = MmsDownloadUiModel.State.AwaitingManualDownload,
                 sizeBytes = 2_048L,
                 expiryTimestamp = 1_700_000_000_000L,
+                isSecondaryUser = false,
```
RankoR, line 32: `isSecondaryUser = true` should be checked as well

RankoR (changes requested):

RankoR:
Looks like secondary user will still see "Tap to download" in the conversation list.

1. ConversationListItemUiMapper resolves the latest message status
2. conversationListMmsDownloadTitleResId() maps IncomingAwaitingManualDownload unconditionally to message_title_manual_download
3. ConversationListItemRow renders that string as the visible snippet
4. ConversationListItemContentDescription also uses it for accessibility

src/com/android/messaging/datamodel/action/DownloadMmsAction.java
```diff
@@ -118,6 +118,11 @@ protected boolean queueAction(final String messageId, final Action processingAct
         final MessageData message = BugleDatabaseOperations.readMessage(db, messageId);
         if (message != null && message.canDownloadMessage()) {
             final Uri notificationUri = message.getSmsMessageUri();
+            if (notificationUri == null) {
+                LogUtil.w(TAG, "DownloadMmsAction: message " + messageId
+                        + " has no notification uri; skipping download");
+                return false;
```
RankoR: This doesn't change message's status, so it will stuck in the current status forever. Better to set a terminal status before returning, like this:

```
updateMessageStatus(null, messageId, message.getConversationId(),
              MessageData.BUGLE_STATUS_INCOMING_DOWNLOAD_FAILED,
              MessageData.RAW_TELEPHONY_STATUS_UNDEFINED);
return true;
```

RankoR (changes requested):
Also, notification still offers a dead-end "Download" action to secondary users

RankoR (approved):

PR #207, opened 2026-07-15
Fix content sliding under the display cutout in landscape

### LINKED ISSUE #174: Certain screens in Compose rewrite need to be inset aware
e.g. use 3-button navigation + camera cutout insets in landscape mode when 
- selecting a conversation for a forwarded message or sharing media to the Messaging app 
- opening conversation settings

m4pl:
Closes https://github.com/GrapheneOS/Messaging/issues/174

Fixes content running under the camera cutout and the navigation bar in landscape on the conversation picker and the message composers, and extracts the shared window inset helpers the other screens were duplicating.

Guarded by a detekt rule that forbids `navigationBarsPadding()`; it needs type resolution, so CI now runs `:app:detektDebug` instead of `:app:detekt`.

### DIFF
--- a/src/com/android/messaging/ui/conversation/screen/ConversationScreenContent.kt
+++ b/src/com/android/messaging/ui/conversation/screen/ConversationScreenContent.kt
@@ -238,7 +238,10 @@ private fun Modifier.conversationScreenContentModifier(
     backdropColor: Color,
 ): Modifier {
     return this
-        .padding(paddingValues = contentPadding)
+        .padding(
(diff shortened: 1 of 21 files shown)


src/com/android/messaging/ui/conversation/screen/ConversationScreenContent.kt
RankoR, line 122: This and `ConversationBlockedBanner.kt:77` are still under the cutout

RankoR (approved):

PR #205, opened 2026-07-14
Fix incoming SMS not imported until the app is opened

m4pl:
Closes https://github.com/GrapheneOS/Messaging/issues/155

Incoming SMS was posted to `ActionServiceImpl`, which `JobScheduler` defers while the app is backgrounded, so messages weren't imported and notifications stayed generic until the app was opened.
`SmsDeliverReceiver` now runs the receive action synchronously inside the `SMS_DELIVER` broadcast window.

PR #229, opened 2026-07-31
Introduce typed id value classes

### LINKED ISSUE #219: Replace raw String identifiers with typed value classes
Replace raw `String` ids with `@JvmInline` `@Serializable` value classes so typed nav keys can't silently swap a conversation id for a message id.

m4pl:
Closes https://github.com/GrapheneOS/Messaging/issues/219

Adds value classes `ConversationId`, `MessageId`, `ParticipantId`, `SubId`. Primitives stay at persistence and Java boundaries via `.value` / `fromOrNull`.

Behavior changes: `ConversationMessageUiModelMapper.map` returns `null` instead of a blank `messageId`, which previously became a duplicate `LazyColumn` key. Some `isEmpty()` checks became `isBlank()`, since the id types only expose blank checks.


m4pl:
Waiting for #205 #204 to be merged

src/com/android/messaging/ui/conversation/messages/mapper/ConversationMessageUiModelMapper.kt
```diff
@@ -16,17 +19,24 @@ import kotlinx.collections.immutable.persistentListOf
...
 internal class ConversationMessageUiModelMapperImpl @Inject constructor(
     private val conversationVCardAttachmentUiModelMapper: ConversationVCardAttachmentUiModelMapper,
 ) : ConversationMessageUiModelMapper {
 
-    override fun map(data: ConversationMessageData): ConversationMessageUiModel {
+    override fun map(data: ConversationMessageData): ConversationMessageUiModel? {
+        val messageId = data.messageId?.takeIf { it.isNotBlank() }
+        val conversationId = data.conversationId?.takeIf { it.isNotBlank() }
+
+        if (messageId == null || conversationId == null) {
```
RankoR, line 33: Maybe log it?

app/src/androidTest/java/com/android/messaging/ui/blockedparticipants/screen/support/BlockedParticipantsTestData.kt
```diff
@@ -9,9 +11,9 @@ import kotlinx.collections.immutable.persistentSetOf
 import kotlinx.collections.immutable.toPersistentList
 import kotlinx.collections.immutable.toPersistentSet
 
-internal const val PARTICIPANT_ID_1 = "p1"
-internal const val PARTICIPANT_ID_2 = "p2"
-internal const val PARTICIPANT_ID_3 = "p3"
+internal val PARTICIPANT_ID_1 = ParticipantId("p1")
```
RankoR, line 14: Should be `participantId1` as it's not `const` anymore

y9san9: Just curious, is it this project convention?

Asking, because in Kotlin we use screaming case for immutable top level data (even if it's not const, and even if it has delegating "get"): https://github.com/JetBrains/kotlin/blob/master/libraries%2Fstdlib%2Fsrc%2Fkotlin%2Fcoroutines%2Fintrinsics%2FIntrinsics.kt#L57

Recently it was added to coding conventions as well: https://kotlinlang.org/docs/coding-conventions.html#property-names

RankoR: @y9san9 oops, I missed that coding conventions have been updated; thanks! @m4pl this means that my comment is no longer relevant :)

src/com/android/messaging/data/conversation/model/draft/ConversationDraft.kt
```diff
@@ -1,14 +1,15 @@
...
+import com.android.messaging.data.conversation.model.ParticipantId
 import kotlinx.collections.immutable.ImmutableList
 import kotlinx.collections.immutable.persistentListOf
 
 @Immutable
 internal data class ConversationDraft(
     val messageText: String = "",
     val subjectText: String = "",
-    val selfParticipantId: String = "",
+    val selfParticipantId: ParticipantId = ParticipantId(""),
```
sdsantos, line 12: Does it make sense if these `ParticipantId("")` would be a constant to add some semantic meaning? Like `ParticipantId.NO_ID` or `ParticipantId.UNDEFINED`?

sdsantos (approved):
Just one small comment regarding the empty value classes.

src/com/android/messaging/data/conversation/model/draft/ConversationDraft.kt
m4pl: I decided to get rid of all the `ConversationId("")` that aren't made for testing. I'd like to avoid creating ParticipantId.UNDEFINED, as that could confuse the logic.

RankoR: To me empty value looks weird as well; Maybe `val selfParticipantId: ParticipantId? = null` then?

m4pl: Now all the empty value classes are only in the tests.

RankoR (approved):

PR #231, opened 2026-08-03
Host conversation list and onboarding in MainActivity

m4pl:
Closes #221

MainActivity becomes the single Compose host with a Navigation 3 back stack, so the conversation list and onboarding are now destinations instead of separate activities.

PR #232, opened 2026-08-03
Flatten conversation navigation into root graph

### LINKED ISSUE #222: Move conversation flow into root graph
Move the conversation screens onto the shared navigation, keeping one activity behind for conversation bubbles to open into.

m4pl:
Closes #222 

Conversation destinations move into the root nav graph, so `ConversationActivity` no longer owns a separate back stack and launch intents are handled by the host.


m4pl:
Waiting for the https://github.com/GrapheneOS/Messaging/pull/231 PR to be merged

RankoR (changes requested):
Widget's "compose" button opens conversation list instead of new message composer

---

Activity recreation after an in-app conversation launch resets user to the conversation list

1. Launch from the launcher
2. Open Archived -> tap a conversation
3. Trigger a recreation not covered by configChanges="orientation|screenSize|keyboardHidden" (e.g. toggle dark mode or change locale)
4. onCreate restores launchGeneration = 1, but getIntent() is the original launcher MAIN intent -> no payload -> launchRequest = null

Probably can be fixed by:

```
if (launchRequest == null || processedLaunchGeneration.value == launchRequest.launchGeneration) {
    return
}
```

in `AppNavGraph.applyLaunchToBackStack`

m4pl:
Both real, both fixed in the follow up branch: [Open New Chat from widget button](https://github.com/m4pl/Messaging/commit/9c51c0366e861d675f5fb16eddf244232b6ff6d7) and [Deliver launch intents as one shot events](https://github.com/m4pl/Messaging/commit/b0777c0f421641c23d4321c94e66e3f7dd679f7c).

The suggested guard would keep the user in the chat when the widget asks to go back to the list, so the fix drops generations instead.

@RankoR

RankoR (approved):

PR #251, opened 2026-08-06
Messaging: add shortcuts for copying phone numbers

Olangu:
Allow phone numbers to be copied from conversation-related UI:

    * Long-press the conversation title to copy the participant's number.
    * Show a participant chooser for group conversations.
    * Copy a sender's number from their message identity area.
    * Tap or long-press the non-action area of the conversation-list avatar popup to copy a number.


thestinger:
It needs a rebase.

PR #265, opened 2026-08-22
Turn onboarding, settings and vCard screens into destinations

m4pl:
Onboarding, archived list, blocked participants, conversation settings and vCard detail become destinations in the root nav graph. Their activities and manifest entries are deleted.
Settings and licenses also get destinations, and their activities are left only for external entry points.


RankoR:
Tapping a contact-card attachment in the Forward or Share picker used to open the in-app vCard viewer. It now fires an external ACTION_VIEW chooser.

Previoisly:

```
ContentType.isVCardType(contentType) -> {
      UIIntents.get().launchVCardDetailActivity(
          context,
          normalizeAttachmentUriForIntent(attachmentUri = attachmentUri),
      )
  }
```

And it's also changed in the test `ConversationAttachmentPreviewEffectTest.kt:135`

RankoR:
Five destinations lose their window title, with no accessibility replacement (activities had `android:label`, now they're gone). We could fix it with `Modifier.semantics { paneTitle = … }` on AppNavDisplay entry wrapper, keyed off the NavKey

src/com/android/messaging/ui/conversationlist/archived/ArchivedConversationListEffectHandler.kt
```diff
@@ -57,7 +60,7 @@ internal class ArchivedConversationListEffectHandlerImpl(
                 DebugUtils.showDebugOptions(activity)
             }
 
-            is Effect.ConversationsUnarchived -> Unit
+            else -> Unit
```
RankoR: Explicit effect branch is better, because `else -> Unit` loses sealed-when exhaustiveness, which can lead to bugs in the future.

RankoR (changes requested):

m4pl:
> Tapping a contact-card attachment in the Forward or Share picker used to open the in-app vCard viewer

Followed Google Messages here. Keeping an activity just for this would make the share flow navigation harder, since the picker is also used by the external share entry point.

@RankoR

RankoR (approved):

PR #271, opened 2026-08-25
Nav entries and legacy cleanup

### LINKED ISSUE #224: Remove legacy intent and activity plumbing
Replace the global intent helper with typed navigation, shrink the system entry points, and delete the old base activities.

m4pl:
Closes #224 

Nav entries now create the view models and pass them into the screens.
Deletes the old Java views and activities nothing uses any more, with their drawables and strings.


RankoR (changes requested):
Looks like `photo_view_activity_title` translation removal is incorrect, it's still used

RankoR (approved):

PR #275, opened 2026-08-26
Add adaptive list detail layout

m4pl:
Adds the material3 adaptive list detail layout, so on wide screens the list and the open conversation sit side by side. Predictive back scales the pane and rounds its corners.

image


src/com/android/messaging/ui/conversation/metadata/delegate/ConversationMetadataDelegate.kt
```diff
@@ -210,14 +219,34 @@ internal class ConversationMetadataDelegateImpl @Inject constructor(
...
+        metadata: ConversationMetadata?,
+    ): Boolean {
+        if (previousMetadata == null) {
+            return false
+        }
+
+        return previousMetadata.isArchived != metadata?.isArchived
+    }
+
+    private suspend fun requestCloseConversation() {
```
RankoR: This can permanently disable closing for the rest of the binding.

`_navigationEvents` is a `MutableSharedFlow(extraBufferCapacity = 1)` with `replay = 0` (:58), and `CollectEvents` collects it from a bare `LaunchedEffect` (`ui/core/CollectEvents.kt:16-18`), so collection is scoped to composition rather than lifecycle. 

If an emission lands while no collector is attached - a configuration change, or the detail pane briefly leaving composition during a scene transition - the event is dropped, but `hasRequestedClose` stays `true`. It is only reset when `conversationId` changes (:91).

After that, for the rest of that conversation binding:
- `onArchiveConversationClick` performs the DB write and emits nothing (:128)
- `confirmDeleteConversation` deletes the conversation and emits nothing (:222)

The second case leaves the user sitting on a conversation that no longer exists.

The latch only exists because the new `hasLeftCurrentList` inference (:99-106) double-fires alongside the explicit archive path. 
If that inference is removed, the latch can go with it.
If it has to stay for now, `requestCloseConversation` should be idempotent by construction rather than by a one-shot flag - routing through `Navigator.closeConversation(conversationId)` gives that for free, since `NavigationReducer.reset` no-ops when the stack is already identical (`NavigationReducer.kt:66`).

src/com/android/messaging/ui/conversation/navigation/ConversationNavEntries.kt
RankoR, line 122: `ConversationScreenRoute.kt:135` maps `CloseConversation -> onNavigateBack()`, which lands here as `appNavigator::back`. 

That's a positional pop, not a conversation-scoped close, and `Navigator` already has the right primitive in `Navigator.kt:64-75`.

`ConversationSettingsNavEntries.kt:43-45` already uses it via onCloseAfterArchive.

With a conversation-scoped child on top - conversation settings, message details, add participants - a CloseConversation for an archived or deleted conversation pops only the child and leaves the user inside a conversation that should be gone.
`closeConversation` drops the whole conversation-owned span.

Repeat calls are harmless (reset no-ops on an identical stack), which removes the need for the hasRequestedClose latch at :241-248.

We could split the callbacks so onNavigateBack keeps meaning "up", and the close event gets its own:

```
onCloseConversation = {
    appNavigator.closeConversation(conversationId = navKey.conversationId)
},
```

src/com/android/messaging/ui/conversationpicker/PickerReviewTransition.kt
```diff
@@ -0,0 +1,80 @@
...
+
+    suspend fun settleTo(isReviewing: Boolean) {
+        if (transitionState.currentState != isReviewing) {
+            transitionState.animateTo(isReviewing)
+        }
+
+        swipeEdge = null
+    }
+
+    suspend fun cancelSeek(totalDurationNanos: Long) {
```
RankoR: This drives `SeekableTransitionState` from outside its own mutator.

`animate` runs the cancel curve here, and each frame `launch`es a *separate* coroutine that calls `transitionState.seekTo(fraction)` (:67-69). 
Each of those acquires the state's internal mutator mutex independently, so frame ordering holds only by main-dispatcher FIFO - nothing guarantees it.
If the user starts a second back gesture while the cancel is still running, that gesture's `seekTo` interleaves with the in-flight cancel frames, and the trailing `snapTo(settledState)` at :73 then fights whatever the new gesture set.

`SeekableTransitionState.animateTo` already does exactly this job - it animates from the current seek fraction and runs inside the state's mutator, so a new gesture preempts it cleanly.

`cancelSeek`, the `totalDurationNanos` plumbing, and `settleScope` in `ConversationPickerScreen.kt:263` can all go.

You can't suspend in the `catch` block - the coroutine is already being cancelled. The usual way around that is to make the cancel a state change and let an effect settle it:

```
// PickerReviewTransition
var isSeeking by mutableStateOf(false)
    private set

suspend fun settleTo(isReviewing: Boolean) {
    transitionState.animateTo(isReviewing)   // unconditional: also unwinds a partialseek
    swipeEdge = null
}

// ConversationPickerScreen
} catch (cancellation: CancellationException) {
    reviewTransition.onSeekCancelled()   // just flips isSeeking = false
    throw cancellation
}

LaunchedEffect(uiState.draft.isReviewing, reviewTransition.isSeeking) {
    if (!reviewTransition.isSeeking) {
        reviewTransition.settleTo(isReviewing = uiState.draft.isReviewing)
    }
}

**Review by RankoR — CHANGES_REQUESTED** [review]

File: src/com/android/messaging/ui/conversationlist/archived/ArchivedConversationListViewModel.kt [hunk 3]
```diff
@@ -87,6 +91,24 @@ internal class ArchivedConversationListViewModel @Inject constructor(
             scope = viewModelScope,
             snapshot = snapshot
         )
+
+        viewModelScope.launch {
+            snapshot
+                .filterNotNull()
+                .map { snapshot -> snapshot.items.isEmpty() }
+                .distinctUntilChanged()
```
RankoR, line 99: This never fires on the unarchive path - which is the primary case this commit is for.

`unarchive()` installs the optimistic override *before* publishing (`:222`), so the list empties while `hasPendingChanges == true`. 
That means the single `false -> true` edge `distinctUntilChanged()` will ever produce arrives inside the guard and is dropped.

`pruneRemovalOverrides` retains a `Removed` override while the conversation is absent from the raw snapshot (`ConversationListOptimisticReducer.kt:96-99`, kept for undo - `prune_removedItemMissingFromRawSnapshot_keepsItForUndo`).
By the time it clears on snackbar dismissal (`:189`), `items.isEmpty()` has been `true`, so `distinctUntilChanged()` has nothing to emit. 
`ConversationListSnapshot` is a data class, so the post-clear publish is `equals`-identical and `StateFlow` conflates it too.

```
after unarchive:          items=0  hasPendingChanges=true
events=[CloseConversation(a)]
after db catch-up:        items=0  hasPendingChanges=true
events=[CloseConversation(a)]
after snackbar dismissed:          hasPendingChanges=false
events=[CloseConversation(a)]
-> CloseArchivedList never emitted
```

Can be fixed with:

```
is Action.UnarchiveSnackbarDismissed -> {
  optimisticSnapshotDelegate.discardRemoval(action.conversationIds)
  if (snapshot.value?.items?.isEmpty() == true) {
    onArchiveEmptied()
  }
}
```

RankoR (changes requested):

src/com/android/messaging/ui/conversationlist/archived/ArchivedConversationListViewModel.kt
m4pl: We only leave when the list empties on its own, for example when you reply from an archived chat and it moves back to the inbox. Leaving right after your own unarchive would close the screen together with the undo snackbar, so that case is skipped.

RankoR (approved):

PR #281, opened 2026-08-29
Fix pre release bugs

RankoR:


src/com/android/messaging/datamodel/MessageNotificationState.java
```diff
@@ -823,17 +862,9 @@ public static void checkFailedMessages() {
...
-                                new NotificationCompat.InboxStyle(builder);
-                        if (line2 != null) {
-                            inboxStyle.addLine(Html.fromHtml(line2.toString()));
-                        }
-                        builder.setStyle(inboxStyle);
-                    } else {
-                        builder.setContentText(line2);
-                    }
 
                     if (builder != null) {
```
m4pl: `builder` is created a few lines up and cannot be null. The block is rewritten anyway, drop the check?

src/com/android/messaging/ui/appsettings/subscription/SubscriptionSettingsViewModel.kt
```diff
@@ -90,6 +111,19 @@ internal class SubscriptionSettingsViewModel @Inject constructor(
         }
     }
 
+    private fun onPhoneNumberConfirmed(phoneNumber: String) {
+        val isValid = phoneNumber.isEmpty() ||
+            PhoneUtils.get(subId.value).getValidSelfE164Number(phoneNumber) != null
```
m4pl: Could this be an injected use case?

src/com/android/messaging/ui/appsettings/subscription/ui/SubscriptionSettingsScreen.kt
```diff
@@ -54,11 +56,11 @@ internal fun SubscriptionSettingsScreen(
     title: String,
     onAction: (Action) -> Unit,
     onNavigateBack: () -> Unit,
+    phoneNumberDialogState: PhoneNumberDialogUiState,
     modifier: Modifier = Modifier,
 ) {
     val scrollBehavior = TopAppBarDefaults.exitUntilCollapsedScrollBehavior()
     var showGroupMmsDialog by remember { mutableStateOf(false) }
```
m4pl: `showGroupMmsDialog` is still local, so one dialog survives rotation and the other does not. Pick one?

src/com/android/messaging/ui/host/SelfPhoneNumberPermissionEffect.kt
```diff
@@ -0,0 +1,49 @@
...
+            ParticipantRefresh.refreshSelfParticipants()
+        }
+    }
+
+    LaunchedEffect(backStack.lastOrNull()) {
+        if (shouldShowOnboarding() || !shouldRequestSelfPhoneNumberPermission()) {
+            return@LaunchedEffect
+        }
+
+        BuglePrefs.getApplicationPrefs().putBoolean(
```
m4pl: `BuglePrefs` and `ParticipantRefresh` straight from the composable, elsewhere this goes through a use case. Not blocking.

src/com/android/messaging/ui/appsettings/subscription/SubscriptionSettingsViewModel.kt
```diff
@@ -80,8 +89,20 @@ internal class SubscriptionSettingsViewModel @Inject constructor(
...
+
+            Action.PhoneNumberDialogDismissed -> {
+                _phoneNumberDialogState.value = PhoneNumberDialogUiState()
+            }
+
+            Action.PhoneNumberEdited -> {
+                _phoneNumberDialogState.update { it.copy(isInvalid = false) }
+            }
+
             is Action.PhoneNumberChanged -> {
```
m4pl: `PhoneNumberConfirmed` would match what this now does?

src/com/android/messaging/ui/appsettings/subscription/ui/SubscriptionSettingsScreen.kt
```diff
@@ -365,10 +368,23 @@ private fun PhoneNumberDialog(
         text = {
             OutlinedTextField(
                 value = phoneNumber,
-                onValueChange = { phoneNumber = it },
+                onValueChange = {
+                    phoneNumber = it
+                    if (isInvalid) {
+                        onNumberEdited()
+                    }
```
m4pl, line 374: This only fires while the error is showing, so the name promises more than it does. `onErrorDismissed`?

src/com/android/messaging/util/PhoneUtils.java
```diff
@@ -503,7 +534,28 @@ private static String getValidE164Number(
...
+        final String country = getSimOrDefaultLocaleCountry();
+        final String validNumber = getValidE164Number(phoneText, country);
+        if (validNumber != null) {
+            return validNumber;
+        }
+
+        final PhoneNumberUtil phoneNumberUtil = PhoneNumberUtil.getInstance();
+        try {
+            final PhoneNumber phoneNumber = phoneNumberUtil.parse(phoneText, country);
+            if (phoneNumberUtil.isPossibleNumber(phoneNumber)) {
```
m4pl, line 558: `isPossibleNumber` accepts a number lifted out of text: `Call me at 5551234567` becomes `+3725551234567` on an EE SIM and `+15551234567` on a US one, and that value goes out as the MMS sender. Reject input with letters? Worth a test either way.

src/com/android/messaging/ui/conversation/messages/delegate/ConversationMessagesDelegate.kt
```diff
@@ -166,6 +172,54 @@ internal class ConversationMessagesDelegateImpl @Inject constructor(
...
+        }
+
+        val durationsByContentUri = coroutineScope {
+            audioContentUris
+                .map { contentUri ->
+                    async { contentUri to resolveAudioDurationMillis(contentUri) }
+                }
+                .awaitAll()
+                .toMap()
+        }
```
sdsantos, line 195: Could it be a good idea to batch these resolves? Is there some parallelism cap somewhere I missed? Or is this a less expensive operation than it seems?

sdsantos (approved):
Just left one comment.

m4pl (approved):

src/com/android/messaging/datamodel/MessageNotificationState.java
```diff
@@ -191,8 +194,7 @@ public MessagingStyle.Message createStyledMessage(Person person) {
             MessagingStyle.Message message =
                     new MessagingStyle.Message(mText, mTimestamp, person);
             if (mAttachmentUri != null && ContentType.isImageType(mAttachmentType)) {
-                message.setData(mAttachmentType,
-                        SharedMemoryImageProvider.Companion.buildUri(mAttachmentUri, mAttachmentType));
+                message.setData(mAttachmentType, mAttachmentUri);
```
inthewaves: The deleted provider decoded the image and recompressed it as PNG before exposing it. That stripped the original container metadata. The replacement puts the original MMS part URI into MessagingStyle. NotificationRecord collects that URI, and NotificationManagerService grants it to each visible NotificationListenerService before invoking `notifyPosted()`. 

An enabled notification listener can now read the original attachment, including EXIF/location or other container data that is not needed to render the notification.

src/com/android/messaging/datamodel/BugleNotifications.java
```diff
@@ -171,11 +173,15 @@ static void createMessageNotification(final String conversationId) {
...
+        // holds every conversation with unseen messages, so it has to be iterated: picking one
+        // entry notifies whichever conversation holds the newest unseen message and silently
+        // drops the rest, and there is no summary notification to surface them.
+        state.mConversationsList.mConversations.stream()
                 .filter(conv -> !isConversationBlocked(conv.mConversationId))
-                .findFirst();
-        conversation.ifPresent(conv -> processAndSend(state, conv));
+                .filter(conv -> !ConversationSnoozeQuery.isConversationSnoozed(
+                        conv.mConversationId))
+                .forEach(conv -> processAndSend(state, conv));
```
inthewaves, line 269: Note that NotificationManagerService enforces maximum of 50 notifications (https://github.com/GrapheneOS/platform_frameworks_base/blob/17/services/core/java/com/android/server/notification/NotificationManagerService.java#L493), maybe the remainder can be in an overflow/summary notification when at >= 51

src/com/android/messaging/datamodel/DatabaseUpgradeHelper.java
```diff
@@ -76,6 +79,53 @@ public int upgradeToVersion3(final SQLiteDatabase db) {
         return 3;
     }
 
+    /**
+     * Purges the rows orphaned while the declared foreign keys went unenforced, parent before
+     * child, so foreign_key_check comes back empty before onConfigure() turns enforcement on. The
+     * view change needs no statement, the views are rebuilt unconditionally above - but the handler
+     * must still exist, or the version check throws and every table is wiped.
```
inthewaves: > foreign_key_check comes back empty before onConfigure() turns enforcement on

SQLiteOpenHelper calls onConfigure() before it begins the upgrade transaction and invokes onUpgrade(), so enforcement is already enabled here

https://github.com/GrapheneOS/platform_frameworks_base/blob/17/core/java/android/database/sqlite/SQLiteOpenHelper.java#L411

The cleanup is needed because enabling foreign keys does not retroactively repair existing violations. This comment could be rewritten to preserve that actual invariant; the current lifecycle description is reversed

inthewaves (changes requested):

src/com/android/messaging/datamodel/MessageNotificationState.java
RankoR: Fixed. BugleNotifications.getNotificationImageUri() decodes the attachment, bakes the EXIF orientation into it and writes a JPEG into the media scratch space.

src/com/android/messaging/datamodel/BugleNotifications.java
RankoR: Good catch, done.

src/com/android/messaging/datamodel/DatabaseUpgradeHelper.java
RankoR: Fixed

src/com/android/messaging/datamodel/BugleNotifications.java
```diff
@@ -600,6 +668,74 @@ public static Bitmap getAvatarBitmap(Context context, Uri avatarUri) {
...
+    static Uri getNotificationImageUri(final Context context, final String partId,
+            final Uri imageUri) {
+        if (partId == null || imageUri == null || !FileProvider.isValidFileId(partId)) {
+            return null;
+        }
+
+        final Uri scratchUri = MediaScratchFileProvider.buildMediaScratchSpaceUri(
+                partId,
+                NOTIFICATION_IMAGE_EXTENSION
+        );
```
inthewaves: This creates a successful scratch space file without a cleanup path. Most of the existing scratch users clean up when ownership ends:

- GIF transcoding deletes its output immediately after reading it
 - Failed/discarded camera, video, and audio captures are deleted
- Rejected or removed draft attachments use deleteTemporaryAttachment()
- Scratch attachments used for MMS retries are destroyed once permanent telephony parts replace them

Notification images don't have an equivalent terminal lifecycle and will accumulate until Android evicts the cache. Could these use a dedicated notification image cache with e.g. bounded age/active-notification-based pruning (e.g. via `NotificationManagerCompat.from(context).getActiveNotifications()`)? Cleanup should probably be rate-limited like run when creating a notification image only if the last sweep is old enough or the cache exceeds a size threshold

Notification images keyed by part ID should also be invalidated when DatabaseHelper.rebuildTables() runs, since old part IDs would no longer be valid. rebuildTables is after an upgrade handler fails or when opening the database with an older schema version (so rare fallback paths and not a normal startup or a successful upgrade)

inthewaves (changes requested):

src/com/android/messaging/datamodel/BugleNotifications.java
inthewaves: Also, while the MMS remains in the telephony database, the cached JPEG is mostly a duplicate, although it has had EXIF removed. But after the message or conversation is deleted, the derived JPEG can remain, retaining the visible image beyond the expected message lifetime.

Internal cache files are app-private and in credential-encrypted storage in Android 17, so unrelated apps cannot directly read them.

Android may evict them, but its documentation explicitly says applications should not rely on automatic cache cleanup. Android app-specific storage guidance (https://developer.android.com/training/data-storage/app-specific#internal-remove-cache)

> Even though Android sometimes deletes cache files on its own, you shouldn't rely on the system to clean up these files for you. You should always maintain your app's cache files within internal storage.

inthewaves: General issue on cache management for scratch space URIs: https://github.com/GrapheneOS/Messaging/issues/284

RankoR: Done. Notification images now live in their own provider and cache directory, swept after every notification pass against `getActiveNotifications()`.

inthewaves (approved):

PR #321, opened 2026-09-16
Fix large conversation load

### LINKED ISSUE #304: v13: App stuck loading messages indefinitely on owner profile in long conversations (~80k messages)
### What happened?

**What happened?**

After updating to v13, opening a conversation with a very large number of old messages (~80,000) causes the app to get stuck. It displays "SMS/MMS, loading messages..." indefinitely, and this loading message then appears in every other conversation as well.

This only happens on the owner profile. Secondary profiles are not affected: they take longer to load the same large conversation compared to smaller ones, but they do eventually load it successfully.

**Steps to reproduce**

1. On the owner profile, open a conversation containing around 80,000 messages.
2. The app gets stuck showing "SMS/MMS, loading messages..." and never finishes loading.
3. Every other conversation opened afterward also shows this stuck loading message.
4. Force-close and reopen the app: everything works normally again, until the large conversation is opened, which triggers the bug again.

**Expected behavior**

The large conversation should load (even if slowly), the same way it does on secondary profiles.

**Additional notes**

- Only reproducible on the owner profile.
- Small/medium conversations load fine on all profiles.
- Secondary profiles load the large conversation correctly, just more slowly than smaller conversations.

### App version

v13

### Logs

_No response_

RankoR:
Fixes https://github.com/GrapheneOS/Messaging/issues/304

The screen loaded the whole conversation in one query and held every message in memory. The query groups on `messages._id` and orders by `received_timestamp`, so SQLite sorted the entire thread into temp b-trees before it could return a single row - 35 s and ~98 MB of Java heap for an 80000-message conversation, with an empty screen the whole time.

Fix:

- `LIMIT` in a subquery over `messages`, before the joins and the `GROUP BY`. A `LIMIT` on the outer statement doesn't help - the sort still runs over the whole conversation first. Bounding the subquery bounds both temp b-trees to the window.
- A `(conversation_id, received_timestamp)` index, schema v5. `index_messages_sort` can't serve this query: `status` sits between those two columns and is filtered with a non-indexable `<>`.
- One window that grows, not pages. Open on the newest 500, double it when the user scrolls within 100 items of the oldest loaded message. It stays one query, one cursor, one observer URI, so invalidation, scroll position and "nothing older left" (a short row count) stay trivial. Paging would have meant replacing that pipeline - the widget, the selection delegate and the photo viewer all read the same URI - for the same user-visible result.

Benchmarks:

| Conversation        | Open -> messages on screen | SQLite query      | Rows loaded  | Java heap         |
| ------------------- | ------------------------- | ----------------- | ------------ | ----------------- |
| **80,000 messages** | **35,453 ms -> 100 ms**    | 35,153 ms -> 32 ms | 80,000 -> 500 | 97.6 MB -> 27.5 MB |
| 300 messages        | 35 ms -> 57 ms             | 13 ms -> 29 ms     | 300          | 26.3 MB -> 26.2 MB |
| 40 messages         | 9 ms -> 20 ms              | 3 ms -> 7 ms       | 40           | 26.2 MB -> 26.2 MB |

Scrolling back through the 80,000-message thread, window doubling each time:

| Window               | 500   | 1,000 | 2,000 | 4,000  |
| -------------------- | ----- | ----- | ----- | ------ |
| Query                | 30 ms | 60 ms | 70 ms | 250 ms |
| Query -> list updated | 18 ms | 47 ms | 39 ms | 73 ms  |

Java heap after flinging back to 4,000 loaded messages: 76 MB - still below what the old code needed just to open the thread.

One-time v4 -> v5 upgrade on that database: 36 ms. The index costs 1.4 MB for 80,000 messages.


src/com/android/messaging/data/conversation/repository/ConversationsRepository.kt
```diff
@@ -51,7 +55,14 @@ internal interface ConversationsRepository {
...
 
-    fun getConversationMessages(conversationId: ConversationId): Flow<List<ConversationMessageData>>
+    /**
+     * Newest-first window over a conversation, re-queried whenever the conversation changes or
+     * [windowSizes] asks for a larger window.
+     */
+    fun getConversationMessages(
+        conversationId: ConversationId,
+        windowSizes: Flow<Int>,
+    ): Flow<ConversationMessagesWindow>
```
inthewaves, line 65: The implementation of this seems to be call `queryConversationMessages()` which traverses `ReversedCursor`, so each returned `messages` list is ordered oldest first, so "Newest-first window" might not be descriptive

src/com/android/messaging/ui/conversation/screen/ConversationScreenContent.kt
```diff
@@ -383,6 +396,35 @@ private fun isScrolledToLatestMessage(listState: LazyListState): Boolean {
...
+        snapshotFlow {
+            val layoutInfo = listState.layoutInfo
+            val lastVisibleItemIndex = layoutInfo.visibleItemsInfo.lastOrNull()?.index
+
+            // Before the first measure pass nothing is visible yet, and asking then would grow the
+            // window on every conversation open.
+            lastVisibleItemIndex != null &&
+                lastVisibleItemIndex >= layoutInfo.totalItemsCount - LOAD_OLDER_MESSAGES_THRESHOLD
+        }
+            .distinctUntilChanged()
```
inthewaves: `snapshotFlow` (https://developer.android.com/reference/kotlin/androidx/compose/runtime/package-summary#snapshotFlow(kotlin.Function0)) is already documented to be similar to that of Flow.distinctUntilChanged)

src/com/android/messaging/widget/WidgetConversationService.java
```diff
@@ -162,20 +165,14 @@ public RemoteViews getViewAt(final int originalPosition) {
...
                             " computed position: " + position +
-                            " scrollToPosition: " + scrollToPosition +
-                            " cursorCount: " + cursorCount +
+                            " cursorCount: " + mCursor.getCount() +
                             " MAX_ITEMS_TO_SHOW: " + MAX_ITEMS_TO_SHOW);
                 }
 
-                intent.putExtra(UIIntents.UI_INTENT_EXTRA_MESSAGE_POSITION, scrollToPosition);
+                intent.putExtra(UIIntents.UI_INTENT_EXTRA_MESSAGE_POSITION, position);
                 if (message.hasAttachments()) {
```
inthewaves, line 176: This position can change between the widget tap and the conversation query. For example, with messages `1…20` ordered oldest to newest, tapping message `10` sends position `10`; if message `21` arrives before the query, `messagePositionToDisplayIndex()` now targets message `11`. The previous oldest-first position remained valid for this append. 

e.g. if `adb shell am kill --user current com.android.messaging.debug` is done with a widget open for a conversation, the conversation widget may no longer receive updates. Tapping on a message in that stale widget after the conversation has received more messages would scroll it to the wrong position. In more natrual occurrences, the Messaging app could be killed by lmkd killing background apps under system memory pressure, which can cause the same receiver-registration loss

I think it would be better to use tapped message ID through the launch request and resolve its position in the loaded window, expanding it if needed

src/com/android/messaging/data/conversation/model/message/ConversationMessagesWindow.kt
```diff
@@ -0,0 +1,8 @@
+package com.android.messaging.data.conversation.model.message
+
+import com.android.messaging.datamodel.data.ConversationMessageData
+
+internal data class ConversationMessagesWindow(
+    val messages: List<ConversationMessageData>,
+    val hasMore: Boolean,
+)
```
inthewaves, line 14: `hasMore` appears to be derived from `ConversationsRepository` code like this:

```kotlin
ConversationMessagesWindow(
    messages = messages,
    hasMore = messages.size >= windowSize,
)
```

However, `messages.size >= windowSize` does not prove that it has more, since it's possible for `messages.size == windowSize`. There would be no more messages, but `hasMore` would still be true, which contradicts the variable name

src/com/android/messaging/datamodel/data/ConversationMessageData.java
```diff
@@ -647,6 +647,49 @@ public static final String getConversationMessagesQuerySql() {
...
+     *
+     * <p>The limit has to be applied to the messages table before the joins and the aggregation.
+     * A LIMIT on the outer statement would not help: the GROUP BY is on messages._id while the
+     * ORDER BY is on messages.received_timestamp, so SQLite sorts every message of the conversation
+     * into a temp b-tree before it can honour the limit, and a conversation with tens of thousands
+     * of messages never finishes loading. Restricting the subquery keeps both temp b-trees down to
+     * the size of the window. Relies on index_messages_conversation_timestamp.
+     *
+     * <p>The limit is inlined rather than bound because a bound argument carries string affinity
+     * here; callers validate it.
```
inthewaves: This comment seems a little misleading, as SQLite documentation permits binding limits even if they're bound as strings: https://www.sqlite.org/lang_select.html#the_limit_clause

AOSP has this exact pattern in ContactsProvider: https://github.com/GrapheneOS/platform_packages_providers_ContactsProvider/blob/17/src/com/android/providers/contacts/picker/ContactsPickerSessionProvider.java#L152-L167

inthewaves (changes requested):

inthewaves:
I think this PR overall is a net improvement since it addresses initial conversation loading while preserving the existing list-based UI. 

For the limitations, it does not bound memory during deep browsing: the window keeps growing. The benchmarks in the PR description would need to be clear whether it's an SMS-onlyl; the fixture with 80000 messages contains only text SMS and would very likely not reflect real-user usage of having a mix of attachment types (audio, image/video, vcards, etc.)

Also each expansion rereads and remaps the expanded window, and subsequent invalidations repeat that work. The code:

1. Queries using the new windowSize (src/com/android/messaging/data/conversation/repository/ConversationsRepository.kt:157).
2. Builds fresh message objects from the entire returned cursor (src/com/android/messaging/data/conversation/repository/ConversationsRepository.kt:511).
3. Maps every message in that window into UI models (src/com/android/messaging/ui/conversation/messages/delegate/ConversationMessagesDelegate.kt:181).

So for example:

  - Open:           query and map   500 messages
  - Expand to 1000: query and map 1,000 messages
  - Total:                       1,500 messages processed
  
Growing from 500 to 1000 messages sends all 1000 through the attachment pipeline again.



It also retains the existing loading/error handling rather than adding separate history-load states and retries.

If e.g. we wanted to also support features such as being able to go to the conversation message position that an attachment came from or searching in messages and being able to jump to the message's position in the conversation, the current window design would still require us to load all messages until the window covers it

In the long term, using Paging 3 could address these limitations with bounded page retention, incremental queries, appropriate refresh keys, and loading UI and would make it easier to support new features such as searching in messages efficiently. Google Messages already uses Paging 3 for both inbox conversation list and the messages within conversations, and Signal also uses paging

src/com/android/messaging/debug/TestDataSeeder.kt
```diff
@@ -2102,3 +2104,84 @@ private fun seedScenarioK(
...
+
+    finalizeConversation(
+        db = db,
+        conversationId = conversationId,
+        latestMessageId = latestMessageId,
+        latestTimestamp = latestTime,
+        snippetText = "Message ${HUGE_CONVERSATION_MESSAGE_COUNT - 1} of " +
+            "$HUGE_CONVERSATION_MESSAGE_COUNT",
+    )
+}
```
inthewaves, line 2187: This only seeds a text-only 80k conversation. If this was used for the benchmarks in the PR, it should be clarified that the benchmark is SMS-only. Actual user conversations would have a mix of attachments (audio, image/video, vcards, etc.) that could affect the benchmark results for both speed and memory usage

src/com/android/messaging/ui/conversation/messages/delegate/ConversationMessagesDelegate.kt
inthewaves, line 193: Growing window design means attachments would be processed again even for messages in previous window. e.g. If initially we query and map 500 messages, in order to expand to a window of 1000, we need to query and map the 500 old messages together with the 500 new messages

The pipeline explicitly maps the complete result, calls `withAudioDurations`, then starts `observeMessagesWithVCardMetadata`

Existing caches appear limited, e.g. 

- [ResolveAudioDurationMillis.kt](https://github.com/RankoR-GOS/Messaging/blob/2b846ba91aac1e15dfc4afc7d92eb9568367ff2d/src/com/android/messaging/domain/media/usecase/ResolveAudioDurationMillis.kt) caches only 256 successful duration results
- The parsed card cache in [VCardEntryRepository.kt](https://github.com/RankoR-GOS/Messaging/blob/2b846ba91aac1e15dfc4afc7d92eb9568367ff2d/src/com/android/messaging/data/vcard/repository/VCardEntryRepository.kt) holds only five URI entries, so larger sets can require repeated parsing. A parsed-card cache hit also does not eliminate embedded-avatar processing: `VCardEntrySummarizer.avatarPhoto()` invokes the downscaler during metadata mapping. Unlike audio's `awaitAll()`, contact cards can initially emit Loading metadata, so this work need not all precede the first message frame

inthewaves (approved):
Looks good especially for initial loading, the other issues can be fixed in the long term with Paging

m4pl (commented):
Looks good. Also checked it on the device.

m4pl (approved):

